PR #11, opened 2021-03-07
Add Firestore database

bbjubjub2494:
# Current state
This contains a working example of getting and displaying data from Google Firebase. Goal: #6

# TODO
- [x] Need to not do this in `MainActivity`, maybe turn it into a unit test
- [x] Can we make a team/organization thingy on the Google side?
- [x] dataclasses

# Points to bring up at Standup meeting
- ~~I need other contributors' google accounts so I can give them admin access.~~
- The current database I made read-only from the app. This is good for some automated tests, but we probably want to add a sandbox database and later a production database.
- We could put dummy datasets and the like in sub-hierarchies
- ~~Martin is working on an issue assigned to me without pinging me.~~

### DIFF
--- a/app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/DatabaseTest.kt
+++ b/app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/DatabaseTest.kt
@@ -1,0 +1,25 @@
+package com.github.HumanLearning2021.HumanLearningApp
+
+import com.github.HumanLearning2021.HumanLearningApp.datamodel.Category
+import com.github.HumanLearning2021.HumanLearningApp.datamodel.Picture
+import com.google.firebase.firestore.ktx.firestore
+import com.google.firebase.firestore.ktx.toObject
+import com.google.firebase.ktx.Firebase
+import junit.framework.TestCase
+
+class DatabaseTest : TestCase() {
+    fun testLoadApple() {
+        Firebase.firestore.document("/databases/demo/pictures/minecraft_apple").get()
+            .continueWithTask {
+                val img = it.result?.toObject<Picture>()
+                assertNotNull(img?.category)
+                img!!.category!!.get()
+            }.continueWith {
+            val category = it.result?.toObject<Category>()
+            assertNotNull(category)
+        }.also {
+            while (!it.isComplete) {
+            }
+        }
+    }
+}
--- a/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/datamodel/Category.kt
+++ b/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/datamodel/Category.kt
@@ -1,0 +1,11 @@
+package com.github.HumanLearning2021.HumanLearningApp.datamodel
+
+/** Plain object for the database entity that represents an abstract category that learners can
+ * learn to recognize and classify.
+ */
+data class Category(
+    /** Name of the category in user language. */
+    val name: String? = null,
+    /** Level 1: presentation, (default) level 2: representation */
+    val comWoorLevel: Int = 1,
+)
--- a/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/datamodel/Picture.kt
+++ b/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/datamodel/Picture.kt
@@ -1,0 +1,12 @@
+package com.github.HumanLearning2021.HumanLearningApp.datamodel
+
+import com.google.firebase.firestore.DocumentReference
+
+/** Plain object for the database entity that represents a picture that learners can classify.
+ */
+data class Picture(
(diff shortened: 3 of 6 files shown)


bbjubjub2494:
FTR publishing `google-services.json` files is OK because:

1. [A Googler said so](https://groups.google.com/g/firebase-talk/c/bamCgTDajkw)
2. It can be reconstructed from the `.apk`, so treating it as secret is pointless and could incur a false sense of security

bbjubjub2494:
>Can we make a team/organization thingy on the Google side?

This is all already in a Google Cloud project, I can add people to it and we only need one so we're good.

bbjubjub2494:
We ought to be able to use Kotlin dataclasses according to the [docs](https://firebase.google.com/docs/firestore/manage-data/add-data#custom_objects)...

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/DatabaseTest.kt
```diff
@@ -0,0 +1,25 @@
+package com.github.HumanLearning2021.HumanLearningApp
+
+import com.github.HumanLearning2021.HumanLearningApp.datamodel.Category
+import com.github.HumanLearning2021.HumanLearningApp.datamodel.Picture
+import com.google.firebase.firestore.ktx.firestore
+import com.google.firebase.firestore.ktx.toObject
+import com.google.firebase.ktx.Firebase
+import junit.framework.TestCase
+
+class DatabaseTest : TestCase() {
```
Nitwix, line 10: Should this really be in app/src/androidTest ? because there is also app/src/test which is maybe more adapted because this test is somewhat independent of android. What do you think?

bbjubjub2494: It requires an authenticated connection to the database. Currently the only way to obtain that is through Android.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/datamodel/Category.kt
```diff
@@ -0,0 +1,11 @@
+package com.github.HumanLearning2021.HumanLearningApp.datamodel
+
+/** Plain object for the database entity that represents an abstract category that learners can
+ * learn to recognize and classify.
+ */
+data class Category(
+    /** Name of the category in user language. */
+    val name: String? = null,
+    /** Level 1: presentation, (default) level 2: representation */
+    val comWoorLevel: Int = 1,
```
Nitwix, line 10: The ComVoor level will not be in the category, it will be set by the administrator for the training session (or maybe tied to the dataset, not sure yet).
also: it's written ComVoor. http://www.susa.be/index.php/services/formations/notre-catalogue/communiquer/142-sensibilisation-a-l-outil-comvoor
(Also: wouldn't an enum be better here ?)

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/datamodel/Picture.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import com.google.firebase.firestore.DocumentReference
+
+/** Plain object for the database entity that represents a picture that learners can classify.
+ */
+data class Picture(
+    /** Location of the image in Cloud Storage. */
+    val url: String? = null,
+    /** The category that the picture depicts. (mandatory) */
+    val category: DocumentReference? = null,
```
Nitwix, line 11: why DocumentReference and not Category object ?

Nitwix (commented):
I put some comments

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/datamodel/Category.kt
Nitwix: you could make some sort of adapter function in the enum maybe. but anyway, it's not gonna be an attribute of Category

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/datamodel/Picture.kt
bbjubjub2494: Then, when we load one image to the database, we would have to load the category as well, and handle duplication and re-serialization. That's cumbersome and inefficient.

bbjubjub2494: Actually I'll walk that back: it's possible to re-serialize nested objects using their primary key (see Firebase docs and blah).

As for efficiency: at my level of abstraction it makes sense to focus on one document at a time, but at the `model.DatabaseInterface` (or whatever it will be called) there will be a specific strategy to efficiently manage `Category` objects so we should be able to accomplish this theret

bbjubjub2494:
superseded

PR #13, opened 2021-03-10
Feature/7 take picture

wattsecs:
Draft pull request with the following features:

- Access the AddPictureActivity from MainActivity for demo purposes
- AddPictureActivity checks for camera permissions and displays an informative message if they are denied or if the devices has no camera
- Live preview if the picture the camera is going to take
- Category selection with visual confirmation

### DIFF
--- a/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivity.kt
+++ b/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivity.kt
@@ -1,0 +1,131 @@
+package com.github.HumanLearning2021.HumanLearningApp
+
+import android.Manifest
+import android.content.pm.PackageManager
+import android.os.Bundle
+import android.view.View
+import android.widget.Button
+import android.widget.Toast
+import androidx.activity.result.ActivityResultLauncher
+import androidx.activity.result.contract.ActivityResultContracts
+import androidx.appcompat.app.AlertDialog
+import androidx.appcompat.app.AppCompatActivity
+import androidx.camera.core.CameraSelector
+import androidx.camera.core.Preview
+import androidx.camera.lifecycle.ProcessCameraProvider
+import androidx.camera.view.PreviewView
+import androidx.core.content.ContextCompat
+import androidx.lifecycle.LifecycleOwner
+import com.google.common.util.concurrent.ListenableFuture
+
+/*
+Activity where an administrator can take a picture to add it to a selected data set
+ */
+class AddPictureActivity : AppCompatActivity() {
+
+    private lateinit var cameraProviderFuture: ListenableFuture<ProcessCameraProvider>
+
+    private val dummyCategories = arrayOf(
+        "category1",
+        "category2",
+        "category3",
+        "category4",
+        "category5",
+        "category6",
+        "category7",
+        "category8",
+        "category9",
+        "category10"
+    )
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        if (cameraIsAvailable()) {
+            when {
+                ContextCompat.checkSelfPermission(
+                    this,
+                    Manifest.permission.CAMERA
+                ) == PackageManager.PERMISSION_GRANTED -> {
+                    setupActivityLayout()
+                }
+                shouldShowRequestPermissionRationale(Manifest.permission.CAMERA) -> {
+                    permissionNeededDialog()
+                }
+                else -> {
+                    setupRequestPermissionLauncher().launch(Manifest.permission.CAMERA)
+                }
+            }
+        } else {
+            permissionNeededDialog()
+        }
+    }
+
+    private fun cameraIsAvailable(): Boolean {
+        return applicationContext.packageManager.hasSystemFeature(PackageManager.FEATURE_CAMERA_ANY)
+    }
+
+    private fun setupRequestPermissionLauncher(): ActivityResultLauncher<String> {
+        return registerForActivityResult(ActivityResultContracts.RequestPermission()) { isGranted: Boolean ->
+            if (isGranted) {
+                setupActivityLayout()
+                Toast.makeText(
+                    applicationContext,
+                    getString(R.string.AddPicture_permissionGrantedToast),
+                    Toast.LENGTH_SHORT
+                ).show()
+            } else {
+                permissionNeededDialog()
+            }
+        }
+    }
+
+    private fun onSelectCategoryButton(v: View) {
+        val builder = AlertDialog.Builder(this)
+        builder.apply {
+            setTitle(getString(R.string.AddPicture_categorySelectionDialogTitle))
+            setItems(dummyCategories) { dialog, category_index ->
+                val button = findViewById<Button>(R.id.selectCategoryButton)
+                button.text = dummyCategories[category_index]
+                button.setBackgroundColor(getColor(R.color.button_set))
+                button.setTextColor(getColor(R.color.black))
+            }
+        }
+        val dialog = builder.create()
+        dialog.show()
+    }
+
+    private fun setupActivityLayout() {
+        setContentView(R.layout.activity_camera)
+        findViewById<Button>(R.id.selectCategoryButton).setOnClickListener(this::onSelectCategoryButton)
+        cameraProviderFuture = ProcessCameraProvider.getInstance(this)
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivity.kt
```diff
@@ -0,0 +1,126 @@
...
+        } else {
+            permissionNeededDialog()
+        }
+    }
+
+    private fun cameraIsAvailable(): Boolean {
+        return applicationContext.packageManager.hasSystemFeature(PackageManager.FEATURE_CAMERA)
+    }
+
+    private fun setupRequestPermissionLauncher() {
```
bbjubjub2494: Would it be possible to override `onRequestPermissionsResult` for that?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivity.kt
```diff
@@ -0,0 +1,126 @@
...
+        requestPermissionLauncher =
+                registerForActivityResult(ActivityResultContracts.RequestPermission()) { isGranted: Boolean ->
```
bbjubjub2494: Could we check that the permission that was granted is actually camera? unless that's redundant

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivity.kt
```diff
@@ -0,0 +1,126 @@
...
+                        permissionNeededDialog()
+                    }
+                }
+    }
+
+    /*
+    Needs to be set on a button
+     */
+    private fun setupOpenLabelSelectionLayoutListener() {
+        openLabelSelectionLayoutListener = View.OnClickListener { it: View ->
```
bbjubjub2494: That listener doesn't need to be lateinit since it's just a lambda. It also doesn't need to be a lambda either. What I like to do with these things is move them to a method an use that method with the `this::onSelectLabelButton`, it's more descriptive.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivity.kt
```diff
@@ -0,0 +1,126 @@
...
+            val builder = AlertDialog.Builder(this)
+
+            builder.apply {
+                setTitle(getString(R.string.AddPicture_labelSelectionDialogTitle))
+                setItems(dummyLabels) { dialog, label_index ->
+                    (it as Button).text = dummyLabels[label_index]
```
bbjubjub2494: This could be excess of caution, but I'm not very confortable with that down-cast. The alternative would be to go do `findViewById`, which is checked at compile time.

```suggestion
                    findViewById<Button>(R.id.selectLabelButton).text = dummyLabels[label_index]
```

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivity.kt
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    private fun bindPreview(cameraProvider: ProcessCameraProvider) {
+        val preview: Preview = Preview.Builder().build()
+        val cameraSelector: CameraSelector =
+                CameraSelector.Builder().requireLensFacing(CameraSelector.LENS_FACING_BACK).build()
+
+        preview.setSurfaceProvider(findViewById<PreviewView>(R.id.cameraPreviewView).surfaceProvider)
+
+        val camera = cameraProvider.bindToLifecycle(this as LifecycleOwner, cameraSelector, preview)
```
bbjubjub2494:
```suggestion
        val camera = cameraProvider.bindToLifecycle(this, cameraSelector, preview)
```
This upcast is redundant.

bbjubjub2494 (changes requested):
Good work! I made some light defensive programming remarks.

I'd like to also draw your attention to a the the [things android-lint noticed on the CI.](https://github.com/HumanLearning2021/HumanLearningApp/pull/13/files#annotation_1138482414) I would also ask that at some point you clean up the commit history with a rebase (or any tool you might use)

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivity.kt
```diff
@@ -0,0 +1,126 @@
...
+    private lateinit var cameraProviderFuture: ListenableFuture<ProcessCameraProvider>
+    private lateinit var requestPermissionLauncher: ActivityResultLauncher<String>
+    private lateinit var openLabelSelectionLayoutListener: View.OnClickListener
+
+    private val dummyLabels = arrayOf("label1", "label2", "label3", "label4", "label5", "label6", "label7", "label8", "label9", "label10")
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setupRequestPermissionLauncher()
+        setupOpenLabelSelectionLayoutListener()
```
Nitwix: I think it's not ideal to make your code depend on state. If later you need to modify something and you accidentally use requestPermissionLauncher before the call to setupRequestPermissionLauncher, it will not work. Maybe the simpler solution would be to make the function return a reference to the newly created object. It also allows you to remove the class attributes.

Nitwix (commented):
Overall it seems good! but just take a look at the comment about state dependency.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivity.kt
```diff
@@ -0,0 +1,126 @@
...
+            }
+            val dialog = builder.create()
+            dialog.show()
+        }
+    }
+
+    private fun setupCameraPreviewView() {
+        setContentView(R.layout.activity_camera)
+        findViewById<Button>(R.id.selectLabelButton).setOnClickListener(openLabelSelectionLayoutListener)
+        cameraProviderFuture = ProcessCameraProvider.getInstance(this)
```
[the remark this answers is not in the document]
wattsecs: This future will be needed to enable picture taking, which is why it has an instance variable.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivity.kt
wattsecs: Yes but as mentioned [here](https://developer.android.com/training/permissions/requesting#request-permission) it would require managing the permission code which is not necessary this way.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivity.kt
wattsecs: The input to RequestPermission is provided at line 49 in `requestPermissionLauncher.launch(Manifest.permission.CAMERA)`

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivity.kt
wattsecs: Agreed

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivity.kt
wattsecs: I agree and implemented it as suggested.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivity.kt
wattsecs: I agree and modified it as suggested.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivity.kt
wattsecs: Yes, fixed

wattsecs:
Commit history was rebased into a single commit.
I am aware of the remaining warnings from android-lint but they were ignored on purpose.

bbjubjub2494 (approved):
All my concerns have been addressed

Nitwix (approved):

PR #16, opened 2021-03-11
design UI Model interface and add dummy Dataset

martin-lenweiter:
First implementation of the part of the model (from the MVP pattern) that provides an interface to access data sets. This includes a dummy data set containing 3 pictures, which can be retrieved by specifying the desired category.


app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/Model/DataSet.kt
```diff
@@ -0,0 +1,40 @@
...
+    val categories: Set<Category>,
+    val comVoorLevel: Int
+) {
+
+    /**
+     * A function to retrieve a picture from the data set given a category
+     *
+     * @param category the category of the image to be retrieved
+     */
+    abstract fun getPicture(category: Category): CategorizedPicture
```
bbjubjub2494:
```suggestion
    abstract fun getPicture(category: Category): Task<CategorizedPicture>
```

This one needs to return an asynchronous task (`com.google.android.gms.tasks.Task`) Dependency is `    implementation 'com.google.android.gms:play-services-tasks:17.2.1'`. (Alternatively, tasks are isomorphic to Kotlin coroutines if you prefer that)

bbjubjub2494 (changes requested):

bbjubjub2494 (approved):

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/Model/DummyDataSetInterface.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    override suspend fun getPicture(categoryString: String): CategorizedPicture {
+        return currentDataSet.getPicture(getCategory(categoryString))
+
+    }
+
+
+    private fun getCategory(name: String):Category {
+        for(category in currentDataSet.categories) {
+            if(category.name.equals(name, ignoreCase = true)) return category
```
Nitwix, line 19: Can't you reuse the equals method you defined in Category?
```suggestion
            if(category.equals(Category(name))) return category
```

Nitwix (changes requested):
Many good things! Maybe just change the name of the DummyUIPresenter to be more precise

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/DummyCategorizedPictureTestActivity.kt
```diff
@@ -0,0 +1,22 @@
...
+import com.github.HumanLearning2021.HumanLearningApp.Presenter.DummyUIPresenter
+import kotlinx.coroutines.launch
+import androidx.lifecycle.lifecycleScope
+
+class DummyCategorizedPictureTestActivity : AppCompatActivity() {
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_dummy_categorized_picture_test)
+
+        val dummyPresenter = DummyUIPresenter()
```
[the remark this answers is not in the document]
bbjubjub2494: The idea is that the Presenter is a big MVP block with more than just datasets concerns in it.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/Model/Category.kt
```diff
@@ -0,0 +1,19 @@
...
+ * @param name the name of the category
+ */
+class Category(val name: String) {
+
+    override fun equals(other: Any?): Boolean {
+        return other is Category && other.name.equals(name, ignoreCase = true)
+    }
+
+    override fun hashCode(): Int {
+        return 17 + 31 * name.hashCode()
```
bbjubjub2494: this hashCode isn't consistent with the case-insensitive `equals`

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/Model/DummyDataSetInterface.kt
bbjubjub2494: Strictly speaking yes, however if Category were to receive more fields later this would break, so it's not very future-proof.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/Model/Category.kt
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Class representing a category, which can be used to classify images
+ *
+ * @param name the name of the category
+ */
+class Category(val name: String) {
+
+    override fun equals(other: Any?): Boolean {
+        return other is Category && other.name.equals(name, ignoreCase = true)
```
bbjubjub2494: It might be desirable to do the [Odersky, Spoon, Venners canEqual](https://www.artima.com/articles/how-to-write-an-equality-method-in-java) here, given that I intend to inherit this class in the Firestore module.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/MainActivity.kt
```diff
@@ -1,11 +1,16 @@
...
 
+import android.content.Intent
 import androidx.appcompat.app.AppCompatActivity
 import android.os.Bundle
 
 class MainActivity : AppCompatActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        setContentView(R.layout.activity_main)
+        setContentView(R.layout.activity_dummy_categorized_picture_test)
```
wattsecs: A button which opens DummyCategorizedPictureTestActivity would make demoing the app easier.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/Model/DataSet.kt
```diff
@@ -0,0 +1,40 @@
...
+ * A class representing a data set of images
+ *
+ * @param name the name of the data set
+ * @param admin the admin who created the data set
+ * @param categories the set of categories of images present in the dataset
+ * @param comVoorLevel the ComVoor level (1-5) describing the difficulty of the data set
+ */
+abstract class DataSet(
+    val name: String,
+    val admin: Admin,
```
wattsecs: Maybe make it a set of admins to allow for collaborative datasets?

NicolasVial: this administrator represents the creator of the dataset, right? If so, I would say there is only one creator. But if another administrator modifies the dataset, he would also be "creator"? And in this case the administrator set makes sense to me.

app/build.gradle
```diff
@@ -75,20 +75,29 @@ connectedCheck {
...
+
     implementation 'androidx.fragment:fragment-ktx:1.3.0'
     implementation 'androidx.activity:activity-ktx:1.2.1'
+
     testImplementation 'junit:junit:4.13.2'
     androidTestImplementation 'androidx.test.ext:junit:1.1.2'
     androidTestImplementation('com.schibsted.spain:barista:3.8.0') {
         exclude group: 'org.jetbrains.kotlin'
     }
+    testImplementation 'org.jetbrains.kotlinx:kotlinx-coroutines-test:1.4.3'
```
bbjubjub2494: duplicate

bbjubjub2494:
The problem is case sensitivity.

```
diff --git a/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/Model/Category.kt b/app/src/main/java/com/github/HumanLearning2021/HumanLearningA>
index f756641..eb40e54 100644
--- a/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/Model/Category.kt
+++ b/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/Model/Category.kt
@@ -6,16 +6,17 @@ package com.github.HumanLearning2021.HumanLearningApp.Model
  *
  * @param name the name of the category
  */
-class Category(val name: String) {
+class Category(name: String) {
+    val name = name.toLowerCase()
 
 
     //TODO: Odersky canEqual
     override fun equals(other: Any?): Boolean {
-        return other is Category && other.name == (name.toLowerCase())
+        return other is Category && other.name == name
     }
 
     override fun hashCode(): Int {
-        return 17 + 31 * name.toLowerCase().hashCode()
+        return 17 + 31 * name.hashCode()
     }

**Review by wattsecs — COMMENTED** [review]
I had the same problems with the tests not passing in cirrus, a rerun solved the problem for me.

**Comment by martin-lenweiter** [comment]
TODO: Make data classes

**Comment by bbjubjub2494** [comment]
> TODO: Make data classes

Problem is: dataclasses are final, and I need to extend them in FireBase.

# ~~Proposal 1~~ if the type system were better
```kotlin
interface CategoryBase {
    abstract val name: String
}

interface DataSetInterface<Category: CategoryBase> {
}
```

And you can implement `DummyCategory` as a dataclass

# Proposal 2
Keep every entity in the interface an `interface`. `DummyCategory` can be a dataclass if desired.
```kotlin
interface Category {
    abstract val name: String
}
interface DataSetInterface {
// blah Category
}
class DummyDataSetInterface {
// blah DummyCategory. You have to downcast arguments but that's the price of abstraction
}
```

martin-lenweiter:
> > TODO: Make data classes
> 
> Problem is: dataclasses are final, and I need to extend them in FireBase.
> # ~Proposal 1~ if the type system were better
> 
> ```kotlin
> interface CategoryBase {
>     abstract val name: String
> }
> 
> interface DataSetInterface<Category: CategoryBase> {
> }
> ```
> 
> And you can implement `DummyCategory` as a dataclass
> # Proposal 2
> 
> Keep every entity in the interface an `interface`. `DummyCategory` can be a dataclass if desired.
> 
> ```kotlin
> interface Category {
>     abstract val name: String
> }
> interface DataSetInterface {
> // blah Category
> }
> class DummyDataSetInterface {
> // blah DummyCategory. You have to downcast arguments but that's the price of abstraction
> }
> ```

I ended up making only the admin class a data class, as the other ones weren't working. As for the category one, this allows to keep case-insensitivity for its name attribute. No need to make category an interface then, right?

bbjubjub2494:
> I ended up making only the admin class a data class, as the other ones weren't working. As for the category one, this allows to keep case-insensitivity for its name attribute. No need to make category an interface then, right?

Your design enforces the "two categories are equal if their names are equal without regards to case" rule and "two images are equal if they belong to the same category". Those are not applicable to the Firebase implementation. Please move this to dummy implementations and keep the interfaces minimal.

bbjubjub2494 (changes requested):
non-minimalistic interface, scheduled to be discussed over voice tomorrow

bbjubjub2494:
The current interface doesn't run counter to any of my requirements. :+1: Also you're right: displayOn doesn't need to be `suspend` as far as I'm concerned.

I'll restate for the record, we still need
- `suspend fun getCategories(): Set<Categories>` (at least for now since the number of categories is tractable)
- `suspend fun putPicture(/* arguments TBD */): Picture`
- `suspend fun putCategory(/* arguments TBD */): Category`

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/Model/DataSetInterface.kt
```diff
@@ -0,0 +1,11 @@
+package com.github.HumanLearning2021.HumanLearningApp.Model
+
+import android.graphics.drawable.Drawable
+
+
+/**
+ * An interface representing the part of the model interacting with data sets
+ */
+interface DataSetInterface {
+    suspend fun getPicture(category: Category): CategorizedPicture
```
bbjubjub2494:
```suggestion
    suspend fun getPicture(category: Category): CategorizedPicture?
```
I realized when experimenting with my implementation: if no picture is available, it might make sense to return null instead of throwing an exception. Compare this to the plural case: if this was "getPictures", we could return an empty list.

app/build.gradle
```diff
@@ -75,20 +75,29 @@ connectedCheck {
 }
 dependencies {
 
+    testImplementation 'org.junit.jupiter:junit-jupiter'
```
bbjubjub2494:
```suggestion
    testImplementation 'org.junit.jupiter:junit-jupiter:5.7.0'
```

Should fix the CI (and also my local build teeheehee)

build.gradle
```diff
@@ -8,6 +8,8 @@ buildscript {
     dependencies {
         classpath "com.android.tools.build:gradle:4.1.2"
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+        classpath "org.jetbrains.kotlin:kotlin-android-extensions:$kotlin_version"
+        testImplementation 'org.junit.jupiter:junit-jupiter:5.7.0'
```
bbjubjub2494:
```suggestion
```
Wrong `build.gradle`. It should go in `app/`. That said there are no tests using JUnit 5 (there were at some point idk why) so you might as well nix that.

martin-lenweiter:
> Many good things! Maybe just change the name of the DummyUIPresenter to be more precise

Agreed. But unless someone has a suggestion now, let's wait until we've discussed the role of the Presenter on friday.

app/build.gradle
bbjubjub2494: this no longer relevant

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/Model/DisplayPictureActivityTest.kt
```diff
@@ -0,0 +1,72 @@
...
+    @After
+    fun cleanUp() {
+        Intents.release()
+    }
+    @Test
+    fun useAppContext() {
+        // Context of the app under test.
+        val appContext = InstrumentationRegistry.getInstrumentation().targetContext
+        assertEquals("com.github.HumanLearning2021.HumanLearningApp", appContext.packageName)
+    }
```
Nitwix: Is this test useful ?
```suggestion

```

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/model/DisplayPictureActivityTest.kt
```diff
@@ -0,0 +1,56 @@
...
+    @After
+    fun cleanUp() {
+        Intents.release()
+    }
+    @Test
+    fun useAppContext() {
+        // Context of the app under test.
+        val appContext = InstrumentationRegistry.getInstrumentation().targetContext
+        assertEquals("com.github.HumanLearning2021.HumanLearningApp", appContext.packageName)
+    }
```
Nitwix: is this useful ? 
```suggestion

```

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/model/DisplayPictureActivityTest.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    @Test
+    fun viewIsDisplayed(){
+        assertDisplayed(R.id.imageView)
```
Nitwix: imageView is not a very descriptive id.
Could you prefix the id with the activity name?
```suggestion
        assertDisplayed(R.id.DisplayPicture_imageView)
```

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/DataSetInterface.kt
```diff
@@ -0,0 +1,9 @@
+package com.github.HumanLearning2021.HumanLearningApp.model
+
+
+/**
+ * An interface representing the part of the model interacting with data sets
+ */
+interface DataSetInterface {
```
Nitwix: Maybe a question of taste, but I wouldn't have put a capital 'S' in Dataset (https://en.wikipedia.org/wiki/Data_set)
(but it's not a big problem)

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/DummyCategorizedPicture.kt
```diff
@@ -0,0 +1,31 @@
...
+        }
+    }
+
+    override fun equals(other: Any?): Boolean {
+        return other is CategorizedPicture && other.category == category
+    }
+
+    override fun hashCode(): Int {
+        return 17 + 31*category.hashCode()
+    }
```
Nitwix: Is it possible to use a data class in order to avoid having to write these?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/DummyCategory.kt
```diff
@@ -0,0 +1,21 @@
...
+class DummyCategory(name: String):Category {
+    override val name = name.toLowerCase()
+
+    override fun equals(other: Any?): Boolean {
+        return other is Category && other.name == name
+    }
+
+    override fun hashCode(): Int {
+        return 17 + 31 * name.hashCode()
+    }
```
Nitwix, line 19: Same: wouldn't a data class do the trick?

Nitwix (approved):
Good improvement from last time. I still left a few comments for general improvements, but in order make the PR accessible as fast as possible, I'll approve.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/DummyCategorizedPicture.kt
martin-lenweiter: yes, thanks. done

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/DummyCategory.kt
martin-lenweiter: Here it's easier not to, as I convert the name argument to lower case before making it a property, to have case-insensitivity between categories

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/DataSetInterface.kt
martin-lenweiter: done

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/model/DisplayPictureActivityTest.kt
martin-lenweiter: my b. I copied this from the ExampleInstrumentedTest and forgot to delete it.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/DummyCategorizedPicture.kt
```diff
@@ -0,0 +1,23 @@
...
+import com.github.HumanLearning2021.HumanLearningApp.R
+import java.lang.IllegalArgumentException
+
+/**
+ * A picture part of the dummy data set. Can be of any of the following categories: "fork", "knife", "spoon"
+ */
+data class DummyCategorizedPicture(override val category: Category) : CategorizedPicture {
+
+    override fun displayOn(activity: Activity, imageView: ImageView) {
+        if(category !is DummyCategory) throw IllegalArgumentException("provide a dummy category to the class constructor")
```
bbjubjub2494, line 14: This invariant can be checked in the primary constructor, but it's good enough right now

bbjubjub2494 (approved):
I will approve this, with **one, very important reservation,** which is that you must use the _squash_ merge button instead of the default one. Ping me if needed. Otherwise LGTM.

PR #25, opened 2021-03-14
Fix cirrus ci crashing on tests from AddPictureActivityTest

wattsecs:
Add a function which deals with system dialogs that open when the camera preview is active


Nitwix (approved):
Seems good to me!

bbjubjub2494 (approved):
LGTM

PR #28, opened 2021-03-15
Bugfix/add picture activity test

Nitwix:
- Adds determinism : the tests are always run in alphabetical order
- Adds wait time before each test: should prevent the following problem : `Waited for the root of the view hierarchy to have window focus and not request layout for 10 seconds.` (yes this is  dirty trick, but I haven't found a better way yet)

### DIFF
--- a/app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivityTest.kt
+++ b/app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivityTest.kt
@@ -3,42 +3,40 @@ package com.github.HumanLearning2021.HumanLearningApp
 import android.Manifest
 import android.content.Intent
 import android.os.Build
-import androidx.test.core.app.ApplicationProvider.getApplicationContext
+import android.view.View
 import androidx.test.espresso.Espresso.onView
+import androidx.test.espresso.UiController
+import androidx.test.espresso.ViewAction
 import androidx.test.espresso.action.ViewActions.click
 import androidx.test.espresso.assertion.ViewAssertions.matches
 import androidx.test.espresso.intent.Intents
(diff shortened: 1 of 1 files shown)


app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivityTest.kt
```diff
@@ -59,6 +81,8 @@ class AddPictureActivityTest {
     @Before
     fun setUp() {
         Intents.init()
+        // this is a dirty trick, but will it work?
```
bbjubjub2494: This is not a helpful comment, please include the answer or reformulate.

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivityTest.kt
```diff
@@ -3,8 +3,12 @@ package com.github.HumanLearning2021.HumanLearningApp
 import android.Manifest
 import android.content.Intent
 import android.os.Build
+import android.util.Log
```
bbjubjub2494: Looks like this import is redundant. Please run "reformat file".

bbjubjub2494 (changes requested):
please rebase and squash

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivityTest.kt
Nitwix: true, I'll reformulate

bbjubjub2494 (approved):
LGTM. crossing my fingers on the CI now.

NicolasVial (approved):
finger crossed !

PR #30, opened 2021-03-16
Update .codeclimate.yml

wattsecs:
Set the threshold of similar lines codeclimate considers to 3

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -2,3 +2,11 @@ exclude_patterns:
 - "**/test/"
 - "**/androidTest/"
 - "**test"
+
+checks:
+  similar-code:
+    config:
+      threshold: 3
+  identical-code:
+    config:
+      threshold: 3


NicolasVial (approved):
I think 3 lines is a good choice for the moment.

bbjubjub2494:
This causes 82 issues to pop up in CodeClimate. the unit of measurement for identical and similar code isn't number of lines, it's [number of AST nodes](https://docs.codeclimate.com/docs/duplication-concept). You need to increase that

wattsecs:
> 
> 
> This causes 82 issues to pop up in CodeClimate. the unit of measurement for identical and similar code isn't number of lines, it's [number of AST nodes](https://docs.codeclimate.com/docs/duplication-concept). You need to increase that

True, sorry about that. I could not find how to set those correctly.
Maybe keeping .codeclimate.yml as it was and marking the codeclimate issues as resolved manually on a case by case basis would be better.

bbjubjub2494:
> Maybe keeping .codeclimate.yml as it was and marking the codeclimate issues as resolved manually on a case by case basis would be better.

I agree. Do you want to close this?

wattsecs:
> 
> 
> > Maybe keeping .codeclimate.yml as it was and marking the codeclimate issues as resolved manually on a case by case basis would be better.
> 
> I agree. Do you want to close this?

Yes

PR #31, opened 2021-03-17
feature/21-Add functionality to add picture, add and retrieve categories to and from dummy dataset

### LINKED ISSUE #21: feature/21-Add functionality to add picture, add and retrieve categories to and from dataset

martin-lenweiter:
add functionality to datasetInterface and Presenter to allow to put pictures to the dataset, and put and get categories. Also add corresponding dummy implementations


bbjubjub2494:
Thought: if you add a Drawable attribute to CategorizedPicture, this might prevent its instances from [serializing](https://developer.android.com/reference/java/io/Serializable), which Nicolas [needs](https://github.com/HumanLearning2021/HumanLearningApp/blob/f69a55b4906eb9d10054cb8424448debfe8d8208/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/DisplayImageActivity.kt#L15).

martin-lenweiter:
> Thought: if you add a Drawable attribute to CategorizedPicture, this might prevent its instances from [serializing](https://developer.android.com/reference/java/io/Serializable), which Nicolas [needs](https://github.com/HumanLearning2021/HumanLearningApp/blob/f69a55b4906eb9d10054cb8424448debfe8d8208/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/DisplayImageActivity.kt#L15).

Good thought. We can discuss this tomorrow the 3 of us. Wouldn't be difficult to change the code to accommodate this.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/presenter/DummyUIPresenter.kt
```diff
@@ -20,4 +21,13 @@ class DummyUIPresenter: UIPresenter {
     override suspend fun getPicture(categoryString: String): CategorizedPicture? {
```
bbjubjub2494: Can we take the opportunity to solve this warning?
```
The corresponding parameter in the supertype 'UIPresenter' is named 'categoryName'. This may cause problems when calling this function with named arguments.
```
```suggestion
    override suspend fun getPicture(categoryName: String): CategorizedPicture? {
```

martin-lenweiter:
I'm thinking that DummyCategory should actually be encapsulated in the DummyDatasetInterface, as this is closer to how the real model is gonna behave. i.e. no one outside the model should be able to create DummyCategory objects. The only way they have access to a such objects is by calling putCategory(). @lourkeur  @NicolasVial what do you two think? Nicolas I see that in your tests you make use of the DummyCategory constructor.

Edit: same applies to DummyCategorizedPicture

bbjubjub2494:
> I'm thinking that DummyCategory should actually be encapsulated in the DummyDatasetInterface, as this is closer to how the real model is gonna behave. i.e. no one outside the model should be able to create DummyCategory objects. The only way they have access to a such objects is by calling putCategory(). @lourkeur @NicolasVial what do you two think? Nicolas I see that in your tests you make use of the DummyCategory constructor.
> 
> Edit: same applies to DummyCategorizedPicture

I think that's defensible from a design perspective, but if it breaks everyone's tests it's unnecessary. Also do you intend to make it nested? private? internal?

martin-lenweiter:
> > I'm thinking that DummyCategory should actually be encapsulated in the DummyDatasetInterface, as this is closer to how the real model is gonna behave. i.e. no one outside the model should be able to create DummyCategory objects. The only way they have access to a such objects is by calling putCategory(). @lourkeur @NicolasVial what do you two think? Nicolas I see that in your tests you make use of the DummyCategory constructor.
> > Edit: same applies to DummyCategorizedPicture
> 
> I think that's defensible from a design perspective, but if it breaks everyone's tests it's unnecessary. Also do you intend to make it nested? private? internal?

Agreed. I was thinking of making them both inner classes (from what I've understood only the parent class is allowed to create instances of inner classes). For now let's keep it as is, but depending on how much we end up using the dummy data set we should consider changing it.

martin-lenweiter:
Make it a draft, as substantial merging work will have to be done once Nicolas and Louis' PR are merged.

Nitwix (approved):
Seems all good to me! Should it still be a draft or is it ready to be merged?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/DummyCategorizedPicture.kt
```diff
@@ -6,18 +6,19 @@ import com.github.HumanLearning2021.HumanLearningApp.R
 import java.lang.IllegalArgumentException
 
 /**
- * A picture part of the dummy data set. Can be of any of the following categories: "fork", "knife", "spoon"
+ * A picture part of the dummy data set. Can be of any of the following categories: "fork",
+ * "knife", "spoon"
  */
 data class DummyCategorizedPicture(override val category: Category) : CategorizedPicture {
 
     override fun displayOn(activity: Activity, imageView: ImageView) {
```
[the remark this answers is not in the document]
martin-lenweiter: It is needed in @lourkeur's implementation, I don't remember exactly why

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/presenter/UIPresenter.kt
```diff
@@ -1,11 +1,35 @@
...
 import com.github.HumanLearning2021.HumanLearningApp.model.CategorizedPicture
+import com.github.HumanLearning2021.HumanLearningApp.model.Category
+import java.io.Serializable
 
 
 /**
- * an interface representing the presenter (in the sense of the MVP design pattern) of the user interface
+ * An interface representing the presenter (in the sense of the MVP design pattern) of the user interface
  */
 interface UIPresenter {
```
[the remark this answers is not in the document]
martin-lenweiter: I didn't do much research on the presenter's role, but actually this seems true. So indeed there's no need for an interface.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/DataSetInterface.kt
```diff
@@ -1,9 +1,58 @@
...
+    /**
+     * A function to retrieve a category
+     *
+     * @param categoryName the name of the desired category
+     * @return the desired category if present, null otherwise
+     */
+    suspend fun getCategory(categoryName: String):Category?
+
+    /**
+     * A function to add a category to the dataset. If the provided name matches that of category
```
bbjubjub2494, line 41: A category's key is not its name in my book, so that constraint is weird to implement for me. It's more natural to create a duplicate category. Obviously the dummy implementation will have problems w/ that though.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/DummyDataSetInterface.kt
```diff
@@ -8,28 +13,51 @@ class DummyDatasetInterface : DatasetInterface {
...
+        if (!categories.contains(category)) throw IllegalArgumentException("The provided category" +
+                "is not present in the dataset")
+
+        for(p in pictures)
+            if(p == DummyCategorizedPicture(category)) return p
+
+        return null
+    }
+
+    override suspend fun putPicture(picture: Serializable, category: Category): CategorizedPicture {
```
bbjubjub2494: What do I do w/ a Serializable? this needs to be a more specific type, or a different one altogether.

EDIT: this applies to the interface, not just this implementation

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/DataSetInterface.kt
martin-lenweiter: Not sure I get what you mean. The name here is just the parameter to the function, so components calling the function might want to get a category that corresponds to a particular name. Or do you want them to refer to categories only by their key, which the `putCategory()` function would return?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/firestore/FirestoreDatasetInterface.kt
```diff
@@ -56,4 +58,16 @@ class FirestoreDatasetInterface(
         val pic = query.get().await().toObjects(PictureSchema::class.java).getOrNull(0)
         return pic?.toPublic()
     }
+
+    override suspend fun putPicture(picture: Serializable, category: Category): CategorizedPicture {
```
bbjubjub2494:
```suggestion
    override suspend fun putPicture(picture: android.net.Uri, category: Category): CategorizedPicture {
```
That's what AddPictureActivity returns and I can work with that.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/DummyDataSetInterface.kt
martin-lenweiter: a Bitmap maybe? or a reference to one?

bbjubjub2494 (changes requested):
just that one thing

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/DataSetInterface.kt
bbjubjub2494: > The name here is just the parameter to the function, so components calling the function might want to get a category that corresponds to a particular name.

There's getCategory for that. They go use that. It does a query. (`SELECT` in SQL parlance). Then they decide if they want to make a new one. `put*` methods are for creating more objects in the database. (`INSERT`)


> Or do you want them to refer to categories only by their key, which the putCategory() function would return?

Ideally yes, otherwise, how do they even know that the given category exists? And what if there's multiple? That's the point of keys.

martin-lenweiter: > > The name here is just the parameter to the function, so components calling the function might want to get a category that corresponds to a particular name.
> 
> There's getCategory for that. They go use that. It does a query. (`SELECT` in SQL parlance). Then they decide if they want to make a new one. `put*` methods are for creating more objects in the database. (`INSERT`)

Aren't we talking about `getCategory()`?

martin-lenweiter: which would correspong to a `SELECT * FROM category WHERE category.name = categoryName` query roughly no?

bbjubjub2494: Yes precisely. Now what I'm complaining about is that currently, putCategory is `SELECT`, then conditionally `INSERT INTO category` based on the wrong assumption that category names should be unique database-wide.

martin-lenweiter: Ah okay misunderstanding then, I didn't get that you were talking about `putCategory`

martin-lenweiter: > Yes precisely. Now what I'm complaining about is that currently, putCategory is `SELECT`, then conditionally `INSERT INTO category` based on the wrong assumption that category names should be unique database-wide.

But then wouldn't it be possible to create a duplicate category with the same parameters (for now name, but we could imagine more), insert it unconditionally into the database with a different key, and return a reference to another category object to the caller?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/DummyDataSetInterface.kt
```diff
@@ -1,5 +1,9 @@
 package com.github.HumanLearning2021.HumanLearningApp.model
 
+import android.graphics.drawable.Drawable
```
bbjubjub2494, line 3: should run code cleanup.

bbjubjub2494 (commented):
Since you severed the connection between `UIPresenter` and `DummyUIPresenter` I have to ask: what interface is Niels (and others) programming against? `UIPresenter`? `DummyUIPresenter`?

martin-lenweiter:
> Since you severed the connection between `UIPresenter` and `DummyUIPresenter` I have to ask: what interface is Niels (and others) programming against? `UIPresenter`? `DummyUIPresenter`?

I actually deleted the interface. Since there will be only one presenter per view, it doesn't make sense to generalize the presenter.

bbjubjub2494:
> > Since you severed the connection between `UIPresenter` and `DummyUIPresenter` I have to ask: what interface is Niels (and others) programming against? `UIPresenter`? `DummyUIPresenter`?
> 
> I actually deleted the interface. Since there will be only one presenter per view, it doesn't make sense to generalize the presenter.

ok gotcha.

bbjubjub2494 (approved):
please squash-merge so that future maintainers don't have to look at this. Otherwise LGTM

PR #34, opened 2021-03-18
Add functionality to main activity for demo purposes

wattsecs:


NicolasVial (approved):
Looks good to me!

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/MainActivity.kt
```diff
@@ -1,22 +1,40 @@
...
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_main)
     }
 
     fun launchToCameraActivity(view: View) {
-        val intent = Intent(this, AddPictureActivity::class.java)
-        startActivity(intent)
+        addPicture.launch(arrayOf("selfie", "yellow head", "some guy"))
```
bbjubjub2494: If we put something like door, sofa, and cat, it matches the virtual room the virtual devices can see. It would be just that bit cooler for the demo wouldn't it?

NicolasVial: That's an awesome idea ! and easy to modify.

bbjubjub2494 (approved):
Either way I built it, it works great, it's going to be good :rocket:

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/MainActivity.kt
wattsecs: On my device it is a face

bbjubjub2494: oh that's odd. If possible taylor it to whoever does the demo

bbjubjub2494: Wait maybe you're using the front camera?

bbjubjub2494: Capture d’écran de 2021-03-18 23-53-48
This is the selfie camera on my Pixel 3 API 29. Is that the same thing you see?

To change camera we have to use the camera app right now I think.

wattsecs: Yeah, I just came to the exact same conclusion. I will go back to using the Pixel 2 Emulator. I changed it sometime this week for some reason.

wattsecs:
Closed because it is no longer needed since the demo is done.

PR #46, opened 2021-03-21
Add google account authentication (#41)

bbjubjub2494:
This is more so geared for a demo.

IMPORTANT: the app now won't build without the passphrase, see discord for instructions.


app/src/main/AndroidManifest.xml
```diff
@@ -14,7 +14,8 @@
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
         android:theme="@style/Theme.HumanLearning">
-        <activity android:name=".DataCreationActivity"/>
+        <activity android:name=".ui.login.GoogleSignInActivity"></activity>
```
Nitwix: does the 'ui' folder have a different purpose than 'view'?

Nitwix (commented):
Just a question about project hierarchy

app/src/main/AndroidManifest.xml
bbjubjub2494: That wasn't my decision, but we're not even using the code that generated this anymore so we might as well name this whatever. Should it be in the top-level like the other activities in your book?

wattsecs (changes requested):
Implementation looks good to me. I would only suggest putting the strings into resources.

bbjubjub2494:
> Implementation looks good to me. I would only suggest putting the strings into resources.

Should be addressed lmk

bbjubjub2494:
ok time for CI rain dance

wattsecs (approved):
My comments have been addressed

Nitwix (approved):
All good for me!

PR #45, opened 2021-03-21
Integrate the dummy dataset with the learning UI

Nitwix:
Also some refactoring because my separation between view and presenter
didn't make much sense.

- [x] I will still need to integrate with #31 once it's merged.
- [x] Also needs a rebase from main (will do once #31 is merged)

PR #48, opened 2021-03-22
Update add picture activity

wattsecs:
Update AddPictureActivity to use Category instead of String
Remove class Category.kt since it is no longer required


app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/AddPictureActivityTest.kt
```diff
@@ -57,12 +58,14 @@ class AddPictureActivityTest {
             }
         }
 
+    private class testCat(override val name: String) : Category
```
bbjubjub2494, line 35: This class is identical to [model.DummyCategory](https://github.com/HumanLearning2021/HumanLearningApp/blob/91a17c46e68112645a37c40469a0b58aaa684c4a/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/DummyCategory.kt) in #45, (and de facto identical in the current main) it might be less redundant to just use that.

wattsecs: Indeed, but since these are tests I wanted them to be as independent as possible from other implementations.

bbjubjub2494: To me, one of the features of dummy implementations is writing tests, but I understand your reasoning. Resolved

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/DataCreationActivity.kt
```diff
@@ -8,6 +8,7 @@ import android.widget.Toast
 import androidx.appcompat.app.AppCompatActivity
 
 import com.github.HumanLearning2021.HumanLearningApp.databinding.ActivityDataCreationBinding
+import com.github.HumanLearning2021.HumanLearningApp.model.Category
```
Jaakik, line 11: Since we are using this we should maybe remove the other category class in " com.github.HumanLearning2021.HumanLearningApp" that was in my previous merge

Jaakik (approved):

Nitwix (approved):
ok for me

PR #49, opened 2021-03-22
firestore: Implement `put*` methods

### LINKED ISSUE #47: Implement put{Picture,Category} for Firestore back-end
follow-up to #21

bbjubjub2494:
Closes #47 

- [x] create and configure the scratch database in Firebase for unit tests
- [x] write unit tests

### DIFF
--- a/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/firestore/FirestoreDatasetInterface.kt
+++ b/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/firestore/FirestoreDatasetInterface.kt
@@ -1,38 +1,56 @@
 package com.github.HumanLearning2021.HumanLearningApp.firestore
 
-import com.github.HumanLearning2021.HumanLearningApp.model.CategorizedPicture
+import android.net.Uri
 import com.github.HumanLearning2021.HumanLearningApp.model.Category
 import com.github.HumanLearning2021.HumanLearningApp.model.DatasetInterface
 import com.google.firebase.FirebaseApp
 import com.google.firebase.firestore.DocumentId
 import com.google.firebase.firestore.DocumentReference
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/firestore/FirestoreDatasetInterface.kt
```diff
@@ -1,38 +1,56 @@
...
 import com.google.firebase.ktx.app
+import com.google.firebase.storage.ktx.storage
 import kotlinx.coroutines.tasks.await
-import java.io.Serializable
+import java.util.*
 
 class FirestoreDatasetInterface(
+    /**
+     * name of a database within the Firebase App
+     */
```
wattsecs, line 19: What do you mean by "within the Firebase App"? Is it not the same as the name give to the dataset (for example "Cooking ustensils")

bbjubjub2494: They are indeed different. My situation is that there can be separate databases in the same Firebase project, like test and production. (Currently we have 2 different test databases, see firebase-setup.md) The "DatasetInterface" needs to select one. The name of a dataset is associated with a dataset, which is an entity within the database. However, we currently operate under the assumption that there is only one dataset in a database. (see #50)

wattsecs: Then I would suggest renaming the class to `FirestoreDatabaseInterface`

wattsecs (approved):
Approved following offline discussion.

martin-lenweiter (approved):
Approved for same reason as Jonas

PR #51, opened 2021-03-23
Add audio feedback in LearningActivity

### LINKED ISSUE #38: Start an audio track upon correct classification of a picture

Nitwix:
On correct classification, the learner hears "bravo!". On incorrect
classification, the learner hears "dommage!".

- [x] I still need to write tests for this added functionnality


app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/LearningAudioFeedback.kt
```diff
@@ -0,0 +1,56 @@
...
+    fun initMediaPlayers(){
+        correctMp = MediaPlayer.create(applicationContext, R.raw.bravo)
+        incorrectMp = MediaPlayer.create(applicationContext, R.raw.dommage)
+    }
+
+    fun releaseMediaPlayers(){
+        // we release when the user pauses the so that we don't eat up resources uselessly
+        correctMp.release()
+        incorrectMp.release()
+    }
```
NicolasVial: I think you forgot a word in the comment right?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/LearningAudioFeedback.kt
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * Gets the MediaPlayer that plays the 'correct' feedback (eg. 'bravo!')
+     * ONLY USE THIS FOR UNIT TESTING
+     */
+    fun __testing_getCorrectMP(): MediaPlayer = correctMp
+    /**
+     * Gets the MediaPlayer that plays the 'incorrect' feedback (eg. 'domamge!')
+     * ONLY USE THIS FOR UNIT TESTING
+     */
+    fun __testing_getIncorrectMP(): MediaPlayer = incorrectMp
```
NicolasVial, line 55: I don't know if it's a good way of doing things to put functions only for testing like this (Even if I don't see another way of doing it). For my knowledge, is this a good way of doing?

NicolasVial (approved):
This is all good for me nice job ! Maybe it would be even better with Tests to check that the MP is launched as we discused but it looks difficult to find a way of doing it. Manual testing is enough for this feature for the moment I think.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/LearningAudioFeedback.kt
Nitwix: yup, was probably tired xD

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/LearningAudioFeedback.kt
Nitwix: yes, this is somewhat bad, but I really didn't see a way to make meaningful tests without it. But the `__testing` prefix at least clearly indicates that it should not be used elsewhere

NicolasVial: okay okay, I think that it is clear enough. Okay for me!

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/LearningActivity.kt
```diff
@@ -21,11 +22,27 @@ class LearningActivity : AppCompatActivity() {
 
     private val dummyPres = DummyUIPresenter(DummyDatabaseService())
     private val learningPresenter = LearningPresenter(DummyDatabaseService())
+    private lateinit var audioFeedback : LearningAudioFeedback
```
wattsecs, line 25: Does it really need to be a lateninit? (could be changed for next sprint)

wattsecs (approved):

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/LearningActivity.kt
Nitwix: I think it does, because I need the applicationContext, but I need to check

PR #50, opened 2021-03-22
Add dummy dataset management functionality

wattsecs:

PR #56, opened 2021-03-24
44: Add ability to choose between presentation and representation mode 

martin-lenweiter:
Only presentation mode is available for now, as representation needs PR #50 to be merged first.


app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/LearningSettingsActivityTest.kt
```diff
@@ -0,0 +1,75 @@
...
+        @After
+        fun cleanUp() {
+            Intents.release()
+            testRule.scenario.close()
+        }
+
+        @Test
+        fun bothButtonsAreDisplayed() {
+            assertDisplayed(R.id.btChoosePresentation)
+            assertDisplayed(R.id.btChooseRepresentation)
```
Nitwix: ok for now, but ideally we need some kind of prefix for the ids, so that we don't end up having conflicts. for example https://github.com/HumanLearning2021/HumanLearningApp/blob/7bbdbf7259a9f7c12aa4352ee5dc69604a75afeb/app/src/main/res/layout/activity_learning.xml#L10

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/LearningSettingsActivityTest.kt
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        @Test
+        fun learningModeTextViewIsDisplayed() {
+            assertDisplayed(R.id.tvMode)
+        }
```
Nitwix: I'm not 100% sure but from what I understand, each time a test case starts, it has to restart the activity and that takes some time. So maybe to speedup the tests it would be better to put all your `assertDisplayed` tests in a same test case (since it essentially tests the same thing)

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/LearningSettingsActivityTest.kt
```diff
@@ -0,0 +1,75 @@
...
+        @Test
+        fun pressingPresentationButtonLaunchesLearningActivity() {
+            clickOn(R.id.btChoosePresentation)
+            intended(hasComponent(LearningActivity::class.java.getName()))
+        }
+
+        @Test
+        fun longClickOnPresentationButtonShowsTooltip() {
+            longClickOn(R.id.btChoosePresentation)
+            //TODO
```
Nitwix: You could use barista's `assertDisplayed("some string")` no?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/Constants.kt
```diff
@@ -0,0 +1,4 @@
+package com.github.HumanLearning2021.HumanLearningApp
+
+
+const val EXTRA_LEARNING_MODE = "EXTRA_LEARNING_MODE"
```
Nitwix: Isn't it best practice to give a name containing the package like ?
`const val EXTRA_MESSAGE = "com.example.myfirstapp.MESSAGE"`
https://developer.android.com/training/basics/firstapp/starting-activity.html

Also, I'm not sure that a Constants class is really a good idea. It will end up being used for all sorts of unrelated things. Better keep it in a companion object in the launching activity

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/LearningSettingsActivity.kt
```diff
@@ -0,0 +1,54 @@
...
+
+        val btPres = findViewById<Button>(R.id.btChoosePresentation)
+        val btRep = findViewById<Button>(R.id.btChooseRepresentation)
+
+        btPres.setOnClickListener {
+            intent.putExtra(EXTRA_LEARNING_MODE, LearningMode.PRESENTATION)
+            startActivity(intent)
+        }
+        btPres.tooltipText = "Play the game in presentation mode: an image must be classified based " +
+                "on an exact copy"
```
Nitwix: use a string resource, it's really what it's made for

app/src/main/res/layout/activity_learning_settings.xml
```diff
@@ -0,0 +1,55 @@
...
+
+        <TextView
+            android:id="@+id/tvMode"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:layout_marginTop="32dp"
+            android:layout_marginBottom="16dp"
+            android:foregroundGravity="center_vertical|center|center_horizontal"
+            android:gravity="center|center_horizontal|center_vertical"
+            android:text="LEARNING MODE"
```
Nitwix: As pointed out by Cirrus, use @string resource

app/src/main/res/layout/activity_learning_settings.xml
```diff
@@ -0,0 +1,55 @@
...
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintTop_toTopOf="parent">
+
+        <Button
+            android:id="@+id/btChoosePresentation"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:layout_margin="8dp"
+            android:layout_weight="1"
+            android:text="Presentation" />
```
Nitwix: Again, use string resource

app/src/main/res/layout/activity_learning_settings.xml
```diff
@@ -0,0 +1,55 @@
...
+
+        <Button
+            android:id="@+id/btChooseRepresentation"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:layout_margin="8dp"
+            android:layout_weight="1"
+            android:text="Representation" />
```
Nitwix: again, use string resource

(It will greatly make our life easier once we start translating the app, EDIT: it's also useful for testing!)

Nitwix (changes requested):
Overall it's good! I like the use of the tooltip!
But please use string resources, it will make our life easier in the future.

martin-lenweiter:
> Overall it's good! I like the use of the tooltip!
> But please use string resources, it will make our life easier in the future.

Yes of course, sorry I forgot about this.

Jaakik (commented):
Looks good to me if Niels suggestions are implemented. For the Jacoco giving 0% coverage locally, try changing the emulator it worked for me last Semester.

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/LearningSettingsActivityTest.kt
martin-lenweiter: just to clarify: what you mean would be something like: `learningSettings_bt_choose_presentation`?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/Constants.kt
martin-lenweiter: Didn't know about this, but yea sounds like a good idea.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/LearningSettingsActivity.kt
martin-lenweiter: done, thanks

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/LearningSettingsActivityTest.kt
```diff
@@ -0,0 +1,70 @@
...
+    fun pressingPresentationButtonLaunchesLearningActivity() {
+        clickOn(R.id.learningSettings_btChoosePresentation)
+        intended(hasComponent(LearningActivity::class.java.name))
+    }
+
+    @Test
+    fun longClickOnPresentationButtonShowsTooltip() {
+        longClickOn(R.id.learningSettings_btChoosePresentation)
+
+        //assertDisplayed("@string/learning_settings_tooltip_presentation")
```
Nitwix: Is it not working?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/LearningSettingsActivity.kt
```diff
@@ -0,0 +1,53 @@
...
+        val intent = Intent(this, LearningActivity::class.java)
+
+        val btPres = findViewById<Button>(R.id.learningSettings_btChoosePresentation)
+        val btRep = findViewById<Button>(R.id.learningSettings_btChooseRepresentation)
+
+        btPres.setOnClickListener {
+            intent.putExtra(EXTRA_LEARNING_MODE, LearningMode.PRESENTATION)
+            startActivity(intent)
+        }
+        btPres.tooltipText = "@string/learning_settings_tooltip_presentation"
```
Nitwix: that's not how this works x) see : https://developer.android.com/reference/android/content/res/Resources#getString(int)

I'll fix it for you

Nitwix (approved):
Ok! My issues with string resources has been adressed.

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/LearningSettingsActivityTest.kt
bbjubjub2494:
```suggestion
        assertDisplayed(R.string.learning_settings_tooltip_presentation)
```

testing as we speak

NicolasVial (approved):
Looks good to me !

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/LearningSettingsActivityTest.kt
```diff
@@ -0,0 +1,69 @@
...
+    fun pressingPresentationButtonLaunchesLearningActivity() {
+        clickOn(R.id.learningSettings_btChoosePresentation)
+        intended(hasComponent(LearningActivity::class.java.name))
+    }
+
+    @Test
+    fun longClickOnPresentationButtonShowsTooltip() {
+        longClickOn(R.id.learningSettings_btChoosePresentation)
+
+        assertDisplayed(R.string.learning_settings_tooltip_presentation)
```
bbjubjub2494: Reverse that, that wasn't the issue

bbjubjub2494:
tried this but it times out. nevermind, keep as is

```kotlin
        testRule.scenario.onActivity {
            assertDisplayed(
                ViewMatchers.hasErrorText(
                    it.getText(R.string.learning_settings_tooltip_presentation).toString()
                )
            )
        }
```

PR #65, opened 2021-03-30
Refactor DataOverviewFragment (now DatasetListFragment) to make it more independent

Nitwix:
I need it to communicate which dataset was selected so the host activity can take appropriate action.

This is a draft because it still needs to be done, but I thought I'd be a good idea to share this modification with the team.


app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/DatasetListFragmentTest.kt
```diff
@@ -1,74 +1,67 @@
...
 class DatasetListFragmentTest {
-
-    @get:Rule
-    val activityScenarioRule: ActivityScenarioRule<AddPictureActivity> = ActivityScenarioRule(
-        Intent(
-            ApplicationProvider.getApplicationContext(),
-            DataOverviewActivity::class.java
-        )
-    )
+    private var dummyDatasets: Set<Dataset>? = null
```
bbjubjub2494:
```suggestion
    private lateinit var dummyDatasets: Set<Dataset>
```

perfect use case for late-initialized vars

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/DatasetListFragmentTest.kt
```diff
@@ -1,74 +1,67 @@
...
+    private var scenario: FragmentScenario<DatasetListFragment>? = null
```
bbjubjub2494:
```suggestion
    private lateinit var scenario: FragmentScenario<DatasetListFragment>
```
same here

bbjubjub2494 (commented):
just a quick tip

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/DatasetListFragmentTest.kt
Nitwix: ah yes, thanks for the tip

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/fragments/DatasetListRecyclerViewAdapter.kt
```diff
@@ -0,0 +1,73 @@
...
+                        findViewById(R.id.DatasetList_elt_im0)
+                    )
+                    catList[1].representativePicture?.displayOn(
+                        hostActivity,
+                        findViewById(R.id.DatasetList_elt_im1)
+                    )
+                    catList[2].representativePicture?.displayOn(
+                        hostActivity,
+                        findViewById(R.id.DatasetList_elt_im2)
+                    )
```
NicolasVial: We will need to have a discussion about the representativePictures. In the DummyDS, no category have a repr. picture which, for me, is a problem. Imo, it would be nice if we "force" to have a representative picture for each category. My idea was : When you add a category to your DS, it launches the camera activity to take a picture which will be the representative of the category, and if a representative picture is deleted from a DS, the first picture we find that belongs in the same category becomes the new representative picture for example. Anyway I think it would be nice to be certain that each category have a representative picture at any point of time to avoid unecessary issues, but we should have a discussion about this I think?

NicolasVial (approved):
Just a little comment about representative pictures, but the rest is really clean for me, good job!
(oh, we still have this little "issue" with the name of "data_overview" and any name that have "data" in it, as you already said it would be probably better with "dataset" since they are the objects we are talking about. But this can be done in future updates ofc.)

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/fragments/DatasetListRecyclerViewAdapter.kt
Nitwix: Yes I agree, it's a bit of a problem to have categories without representatives. Let's discuss this tomorrow maybe

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/fragments/DatasetListRecyclerViewAdapter.kt
```diff
@@ -0,0 +1,91 @@
...
+        val ds = datasetList.value?.get(position)
+        holder.itemView.apply {
+            if (ds != null) {
+                findViewById<TextView>(R.id.DatasetList_elt_text).text = ds.name
+
+                setRepresentativePictures(this, ds)
+            }
+        }
+        // so that the holder can call the callback with the correct dataset
+        holder.bindDataset(ds!!) // should be safe because position has to be  in {0..size-1}
```
bbjubjub2494: comment is a bit confusing given that the previous bit of code was null-checked. If we assume that datasetList.value isn't null you might as well put a double bang at line 54. Although with that said it still looks like a race condition with line 36... Maybe you meant to alias `mutableDatasetList` from the enclosing fragment?

Nitwix: Indeed there are some problems in this part. Will look into it

Nitwix: I think I found a good solution, will push soon

app/src/main/AndroidManifest.xml
```diff
@@ -14,14 +14,15 @@
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
         android:theme="@style/Theme.HumanLearning">
-        <activity android:name=".view.DisplayImageSetActivity"/>
-        <activity android:name=".DataOverviewActivity"/>
+        <activity android:name=".view.DisplayImageSetActivity" />
+        <activity android:name=".DataOverviewActivity"
+            android:exported="true"/><!-- TODO : remove, only for testing -->
```
bbjubjub2494: Can this be removed yet?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/CategorizedPicture.kt
```diff
@@ -1,6 +1,7 @@
 package com.github.HumanLearning2021.HumanLearningApp.model
 
 import android.app.Activity
+import android.content.Context
```
bbjubjub2494: That's redundant is it not?

bbjubjub2494 (approved):
I think we're done with concurrency problems. I like your solution. Just 2 optional nitpicks, otherwise LGTM.

app/src/main/AndroidManifest.xml
Nitwix: yep, will do

PR #84, opened 2021-04-13
Fix tests

bbjubjub2494:
Demo data got updated behind the scenes.


wattsecs (approved):
I agree with the changes.

Nitwix (approved):
Yep, I should've made a new database, it would have been simpler. Maybe in the future we can make a unittest database?

bbjubjub2494:
> Yep, I should've made a new database, it would have been simpler. Maybe in the future we can make a unittest database?

My vision at first was that both scratch and demo were primarily unit test databases. This can be discussed however. I think the main cause here was the closed-world assumption on the part of the unit test, which I didn't really fix here. We could also not add any new apples or change the number, that would work too.

PR #88, opened 2021-04-15
Cleanup the project hierarchy by moving and renaming some files

Nitwix:
This is (was) a draft because:
- [x] Need to rename a few other things
- [x] Needs to be merged last so that it doesn't cause too many conflicts

Advice for reviewing this : look at this branch on android-studio, because the "Files changed" doesn't make the project hierarchy very clear.


app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_list_fragment/GoogleSignInFragmentTest.kt
```diff
@@ -1,12 +1,10 @@
-package com.github.HumanLearning2021.HumanLearningApp
+package com.github.HumanLearning2021.HumanLearningApp.view.dataset_list_fragment
```
bbjubjub2494: that doesn't seem to fit

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/GoogleSignInFragment.kt
```diff
@@ -11,11 +11,8 @@ import android.widget.Button
 import android.widget.TextView
```
bbjubjub2494, line 11: Should this fragment follow its test?

bbjubjub2494 (commented):
I'd like to see a rebased version before I approve, but I like what I'm seeing

maybe avoid running reformat, that can cause unnecessary merge conflicts (altho long term we should format at some point

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_list_fragment/GoogleSignInFragmentTest.kt
Nitwix: yup, it was a mistake, it's fixed

wattsecs (approved):
The suggested restructuration makes sense to me

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/GoogleSignInFragment.kt
Nitwix: I'm not really sure what you mean by this, but do you have a suggestion for where to put this fragment? I wasn't sure where it would fit best

bbjubjub2494: I was thinking most things in the `fragments` package are really widgets, especially given that activities are now going to be fragments so I would call them that and run some renames...

bbjubjub2494: Oh and my original comment was about this class not being in the same package as its `-Test' but that's fixed

Nitwix: Is it ok if I let you do the renaming? I'm not really sure how you would do the renaming. I already deleted the fragments package because it didn't make sense to keep it since we will turn everything into fragments. I think it's best if we make packages by logical units (like I did with the dataset_list_fragment).

If you want to do the renamings, I won't push anything until you've made your changes so that we don't have conflicts. Ok for you?

bbjubjub2494: It's just doing smart renames in Studio but I'll do it now

bbjubjub2494: here you go!

Nitwix:
I'm going to rebase on main, don't push please.

NicolasVial (approved):
Looks good to me, good changes.

PR #99, opened 2021-04-16
[cleanup] address or mark compilation warnings

bbjubjub2494:
This addresses all warnings seen when building the App, save for two:

- `ExperimentalStdlibAPI` things. That should resolve itself over time when it's no longer experimental.
- A couple codegen warnings around Room. We can deal with that in Room-related PRs.


Nitwix (approved):
Good initiative!

NicolasVial (approved):
Good cleanup. Thank you !

PR #100, opened 2021-04-16
Improve picture management

### LINKED ISSUE #97: Feature request : DatabaseManagement::getRandomPicture(category : Category): CategorizedPicture?
description: this method retrieves one random picture from the database, that is in the given category.

Use case : This method would allow the learning activity to not have to download all pictures in a category just to show the next picture to sort.

Maybe it will also come in handy in other places, I don't know

wattsecs:
Adds support for picture ids to allow for better loading of and management of pictures

TODO:
- [x]  Fix DisplayImageActivityTest
- [x]  Fix DipslayImageSetActivityTest


NicolasVial (approved):
good fixes and useful new management functions thank you !

bbjubjub2494 (commented):
What is the reasoning for adding an `id` field to Firestore pictures instead of using the tail of the path like with categories?

wattsecs:
> 
> 
> What is the reasoning for adding an `id` field to Firestore pictures instead of using the tail of the path like with categories?

~~I don't understand what you mean, Categories have an id don't they?~~ Nervermind... I will fix this

bbjubjub2494:
> > What is the reasoning for adding an `id` field to Firestore pictures instead of using the tail of the path like with categories?
> 
> ~I don't understand what you mean, Categories have an id don't they?~ Nervermind... I will fix this

This might be worth upstreaming into FirestoreDocument honestly

bbjubjub2494 (approved):
LGTM

PR #104, opened 2021-04-21
Upload picture from local storage (#94)

bbjubjub2494:
created a separate activity to avoid starting the Camera.

### DIFF
--- a/app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/DatasetsOverviewActivityTest.kt
+++ b/app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/DatasetsOverviewActivityTest.kt
@@ -42,7 +42,7 @@ class DatasetsOverviewActivityTest {
     @get:Rule
     var hiltRule = HiltAndroidRule(this)
     @get:Rule
-    val activityScenarioRule: ActivityScenarioRule<TakePictureActivity> = ActivityScenarioRule(
+    val activityScenarioRule: ActivityScenarioRule<DatasetsOverviewActivity> = ActivityScenarioRule(
         Intent(
             ApplicationProvider.getApplicationContext(),
             DatasetsOverviewActivity::class.java
--- a/app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/DatasetsOverviewActivityTest.kt
+++ b/app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/DatasetsOverviewActivityTest.kt
@@ -42,7 +42,7 @@ class DatasetsOverviewActivityTest {
     @get:Rule
     var hiltRule = HiltAndroidRule(this)
     @get:Rule
-    val activityScenarioRule: ActivityScenarioRule<AddPictureActivity> = ActivityScenarioRule(
+    val activityScenarioRule: ActivityScenarioRule<TakePictureActivity> = ActivityScenarioRule(
         Intent(
             ApplicationProvider.getApplicationContext(),
             DatasetsOverviewActivity::class.java
--- a/app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/DisplayDatasetActivityTest.kt
+++ b/app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/DisplayDatasetActivityTest.kt
@@ -149,6 +149,7 @@ class DisplayDatasetActivityTest {
     fun clickOnMenuAddNewPictureWorks() {
         openActionBarOverflowOrOptionsMenu(getInstrumentation().targetContext)
         onView(withText(R.string.add_new_picture)).perform(click())
+        onView(withText(R.string.use_camera)).perform(click())
 
         intended(
             allOf(
@@ -158,6 +159,7 @@ class DisplayDatasetActivityTest {
         )
 
         Espresso.pressBack()
+        Espresso.pressBack()
 
         categories = dataset.categories
         assumeTrue(categories.isNotEmpty())
@@ -166,6 +168,7 @@ class DisplayDatasetActivityTest {
 
             openActionBarOverflowOrOptionsMenu(getInstrumentation().targetContext)
             onView(withText(R.string.add_new_picture)).perform(click())
+            onView(withText(R.string.use_camera)).perform(click())
 
             onView(withId(R.id.takePictureButton)).perform(click())
             onView(withId(R.id.selectCategoryButton)).perform(click())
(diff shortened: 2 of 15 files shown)


app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/DisplayDatasetActivityTest.kt
```diff
@@ -149,6 +149,7 @@ class DisplayDatasetActivityTest {
     fun clickOnMenuAddNewPictureWorks() {
         openActionBarOverflowOrOptionsMenu(getInstrumentation().targetContext)
         onView(withText(R.string.add_new_picture)).perform(click())
+        onView(withText("Use camera")).perform(click())
```
wattsecs: I suggest using a resoruce as it above because it makes everything more flexible (e.g for translations etc)

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/DisplayDatasetActivityTest.kt
```diff
@@ -166,6 +168,7 @@ class DisplayDatasetActivityTest {
 
             openActionBarOverflowOrOptionsMenu(getInstrumentation().targetContext)
             onView(withText(R.string.add_new_picture)).perform(click())
+            onView(withText("Use camera")).perform(click())
```
wattsecs: I suggest using a resoruce as it above because it makes everything more flexible (e.g for translations etc)

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/DatasetsOverviewActivityTest.kt
```diff
@@ -42,7 +42,7 @@ class DatasetsOverviewActivityTest {
     @get:Rule
     var hiltRule = HiltAndroidRule(this)
     @get:Rule
-    val activityScenarioRule: ActivityScenarioRule<AddPictureActivity> = ActivityScenarioRule(
+    val activityScenarioRule: ActivityScenarioRule<TakePictureActivity> = ActivityScenarioRule(
         Intent(
             ApplicationProvider.getApplicationContext(),
             DatasetsOverviewActivity::class.java
```
NicolasVial: It is probably my bad from previous PRs but I think the activityScenarioRule should be ActivityScenarioRule DatasetsOverviewActivity instead of TakePictureActivity no? Or am I missing something?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/SelectPictureActivity.kt
```diff
@@ -0,0 +1,97 @@
...
+        Glide.with(this).load(pic).into(findViewById(R.id.selectedPicturePreview))
+    }
+
+    private fun launchOpenPicture() {
+        val intent = Intent(Intent.ACTION_OPEN_DOCUMENT).apply {
+            addCategory(Intent.CATEGORY_OPENABLE)
+            type = "image/*"
+
+        }
+        @Suppress("DEPRECATION")  // FIXME: use something non-deprecated
```
wattsecs, line 61: If I remember correctly activity contracts are what should be used now

app/src/main/res/layout/activity_add_picture.xml
```diff
@@ -0,0 +1,29 @@
...
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    tools:context=".view.dataset_editing.AddPictureActivity">
+
+    <Button
+        android:id="@+id/select_existing_picture"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="224dp"
+        android:text="Select Existing Picture"
```
wattsecs: Use a resource

app/src/main/res/layout/activity_add_picture.xml
```diff
@@ -0,0 +1,29 @@
...
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent" />
+
+    <Button
+        android:id="@+id/use_camera"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:text="Use camera"
```
wattsecs: Use a resource

app/src/main/res/layout/activity_select_picture.xml
```diff
@@ -0,0 +1,75 @@
...
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@id/selectPictureGuideline1"
+        tools:ignore="ButtonStyle" />
+
+    <Button
+        android:id="@+id/choosePictureButton"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:text="Choose Picture"
```
wattsecs: Use a resource

wattsecs (changes requested):
Small comments but otherwise all good

NicolasVial (approved):
Looks good to me, just 1 comment about something that is probably my bad, and maybe use String ressources in .xml files as suggested by the CI. For the FIXME, see Jonas suggestion.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/SelectPictureActivity.kt
bbjubjub2494: yes, but the intent in question does not support contracts according to the [documentation](https://developer.android.com/training/data-storage/shared/documents-files#open-file) so there's nothing we can do until the Android API updates itself. Unless I got that wrong.

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/DatasetsOverviewActivityTest.kt
bbjubjub2494: that's an old mistake but you're correct, it's the wrong type. I'll fix that now.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/SelectPictureActivity.kt
wattsecs: Indeed, in that case I wouldn't know a better way

wattsecs (approved):
My concerns have been addressed

PR #107, opened 2021-04-22
Feature/86 Basic Search

Jaakik:
This is Basic Implementation of Search Activity by Dataset Names, for now UI and Search are done In the same activity but it would be better to use a presenter for more complicated scenarios. Just Now I realised that it might be better to adapt DatasetOverview for the search by Name, I will adapt the code in that direction

### DIFF
--- a/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/Search.kt
+++ b/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/Search.kt
@@ -65,6 +65,5 @@ class Search : AppCompatActivity() {
         })
     }
 }
-    }
 
-}
+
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/Search.kt
```diff
@@ -0,0 +1,69 @@
...
+        adapter = ArrayAdapter<String>(this, android.R.layout.simple_list_item_1, datasetList)
+        listView.adapter = adapter
+        searchView.setOnQueryTextListener(object : SearchView.OnQueryTextListener {
+            override fun onQueryTextSubmit(query: String): Boolean {
+                if (list.contains(query)) {
+                    adapter.filter.filter(query)
+                } else {
+                    Toast.makeText(this@Search, "No Match found", Toast.LENGTH_LONG).show()
+                }
+                return false
```
martin-lenweiter, line 59: why do you always return false?

martin-lenweiter (approved):

PR #112, opened 2021-04-22
Use dummy database instead of demo2 for demo

Nitwix:
because firestore quota exceeded

also other fixes for the demo


wattsecs (approved):

wattsecs (approved):
I approve these changes, mainly since they are necessary for the demo

Nitwix:
This PR should probably be closed, as it doesn't bring anything useful anymore. What do you think?

PR #117, opened 2021-04-23
DummyCategorizedPicture: use glide

bbjubjub2494:
that makes displayOn robust, in turn preventing the UI test from breaking.


Nitwix (approved):
Sorry I didn't see earlier

martin-lenweiter (approved):

PR #128, opened 2021-04-30
statically enforce that picture/category/dataset ids are strings

bbjubjub2494:
We make this assumption in multiple places and we agree to it at the design stage now, so it makes sense to express it in code


wattsecs (approved):
I agree

bbjubjub2494:
(missed a couple `as String`)

Nitwix (approved):
good idea

PR #139, opened 2021-05-05
[housekeeping] add demo pitch

wattsecs:
Add a markdown file containing the demonstration pitch. Every subsequent demo should base itself on it and improve it such as to have a good pitch ready for the final demonstration.

### DIFF
--- a/hl_pitch.md
+++ b/hl_pitch.md
@@ -3,7 +3,8 @@
 ## Preparation
 
 Do the path in the application at least once before to see if anything crashes.
-Predownload an image if the demo is done on an emulator. Then add this image during the demo instead of taking one.
+
+Predownload an image if the demo is done on an emulator, then add this image during the demo instead of taking one.
 
 ## Script
 
--- a/hl_pitch.md
+++ b/hl_pitch.md
@@ -15,14 +15,14 @@ The problem now is that the same people do not have good abstraction capabilites
 A method to teach this abstraction is picture matching, which would traditionally require printing lots of images. With our app you can do all of this digitally, which greatly simplifies the process.
 So let's learn!
 
-I have a tablet here with a big screen so I can see the images better. Let's see what datasets are available to me...
+Let's see what datasets are available to me...
 
-_goes and looks at dataset overview_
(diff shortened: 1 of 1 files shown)


NicolasVial (approved):
Good for now, Thank you (Do not merge it now please, I will modify some stuff for this week demo :) )

hl_pitch.md
```diff
@@ -0,0 +1,28 @@
...
+
+_goes and looks at dataset overview_
+
+But there are only very few images... I want to have more objects to match! This is why I am going to go and augment my dataset.
+
+_adds picture and shows it added afterwards_
+
+Now let's learn some abstraction!
+
+_goes and learns in representation mode_
```
Nitwix: Shouldn't we put the learning before the editing, to first showcase the main feature of the app?

wattsecs: I decided the take the approach "first let's look what I can learn on" but one could argue that "I want to learn, oh my experience was not that good let's add more images" might actually make more sense

bbjubjub2494: my iteration of the demo was criticised with this exact angle

NicolasVial:
```suggestion
So let's learn!

Let's see what datasets are available to me...

_goes and looks at datasets overview (In the learning part)_ 

_goes in representation mode and start learning and remark that a picture shouldn't be in this dataset_

 This picture shouldn't be in this dataset and I don't have enough pictures to learn well... Let's see what is going on.
 
 _goes in datasets overview, select the good dataset, first add a new picture from the storage, then go check to see if the picture has been added. There is also the picture that should not be in this dataset, delete it._
 
 Nice, now it is perfect ! I can go back to learning with my improved dataset.
```

NicolasVial (commented):
What do you guys think about this idea ? open to any comment :)

hl_pitch.md
wattsecs: `Let's select the representation mode...` is not necessary imo. We can see that you do it.

NicolasVial: You're right :)

bbjubjub2494 (approved):

PR #140, opened 2021-05-05
Feature/123 make learning clearer

### LINKED ISSUE #123: Improve learning UI to make usage clearer

Nitwix:
Wraps images from learning UI inside a card view to have rounded corners and a border that indicates with which element to interact.

The idea was that first the image to sort has a green border to indicate that you should interact with it. Then, once you start dragging it, the target categories have the green border, to indicate you should drag the image on one of the targets. If it is wrong, the border of the category turns red to indicate a mistake.

This could not be implemented yet because of a lack of time on my part. I'll add an issue so that we can implement this in a future sprint.

### DIFF
--- a/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningFragment.kt
+++ b/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningFragment.kt
@@ -9,6 +9,7 @@ import android.util.Log
 import android.view.*
 import android.widget.ImageView
 import androidx.activity.OnBackPressedCallback
+import androidx.cardview.widget.CardView
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.lifecycleScope
 import androidx.navigation.fragment.findNavController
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningFragment.kt
```diff
@@ -222,29 +227,29 @@ class LearningFragment: Fragment() {
         return true
     }
 
+    private fun onImageToSortTouched(view: View, event: MotionEvent): Boolean {
+        val clipDataLabel = "My Clip Data"
```
wattsecs: Might not be very useful, but for the sake of completness maybe make it related to the actual data?

wattsecs (approved):
One very minor comment, otherwise all good

Nitwix:
It seems there is a compilation error. Fix in progress

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningFragment.kt
Nitwix: yep. this still dates from the prototyping phase of this fragment. Will fix

NicolasVial (approved):
Way clearer, good job.

PR #153, opened 2021-05-09
Add save actions plugin

Nitwix:
This plugin adds the ability to reformat + optimize imports on save, which seems useful in my opinion.

I don't know if everyone would be ok with installing this plugin, let me know.


bbjubjub2494:
tested on my machine. I have to comment that I had to manually install the plugin for it to do anything. Arguably that could be spun as being an opt-in mechanism, but then we should probably document that for contributors. Other than that, I had no issue installing the plugin and I'm willing to use it from now on.

wattsecs (approved):
Tested locally

bbjubjub2494 (approved):
good enough

PR #154, opened 2021-05-09
Feature/92 learning with arbitrary nb of categories

### LINKED ISSUE #92: Adapt learning to allow an arbitrary number of categories

Nitwix:
Closes #92 

- [x] I still have to adapt the tests (sorry for the delay)

### DIFF
--- a/app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningTest.kt
+++ b/app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningTest.kt
@@ -158,7 +158,7 @@ class LearningTest {
         val NUMBER_OF_ATTEMPTS = 50
         var foundImageChange = false
         // tries drag and drop NUMBER_OF_ATTEMPTS times until there is a change in the image to sort
-        // (change in category)
+        // (change in description)
         // prob to have the same color NUMBER_OF_ATTEMPTS (eg 100) times in a row with a working UI is
         // P = (1/NUMBER_OF_CATEGORIES)^NUMBER_OF_ATTEMPTS (eg (1/3)^100, aka very small)
         for (i in 1..NUMBER_OF_ATTEMPTS) {
--- a/app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningTest.kt
+++ b/app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningTest.kt
@@ -13,14 +13,14 @@ import com.github.HumanLearning2021.HumanLearningApp.R
 import com.github.HumanLearning2021.HumanLearningApp.TestUtils
 import com.github.HumanLearning2021.HumanLearningApp.hilt.DatabaseManagementModule
 import com.github.HumanLearning2021.HumanLearningApp.hilt.Demo2Database
-import com.github.HumanLearning2021.HumanLearningApp.model.DatabaseManagement
-import com.github.HumanLearning2021.HumanLearningApp.model.DefaultDatabaseManagement
-import com.github.HumanLearning2021.HumanLearningApp.model.DummyDatabaseService
+import com.github.HumanLearning2021.HumanLearningApp.model.*
 import com.schibsted.spain.barista.assertion.BaristaVisibilityAssertions.assertDisplayed
+import com.schibsted.spain.barista.assertion.BaristaVisibilityAssertions.assertNotDisplayed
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningFragment.kt
```diff
@@ -60,13 +57,58 @@ class LearningFragment : Fragment() {
...
+                    listOf(learningCat1, learningCat2)
+                }
+                else -> listOf(learningCat0, learningCat1, learningCat2)
+            }
+        }
+    }
+
+    /**
+     * TODO
+     */
```
NicolasVial: This TODO (and others similar to this one) is about the KDoc not the content of the method right?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/presenter/LearningPresenter.kt
```diff
@@ -3,73 +3,91 @@ package com.github.HumanLearning2021.HumanLearningApp.presenter
...
+     */
+    suspend fun updateForNextSorting(
+        activity: Activity,
+        targetViews: List<ImageView>,
+        sourceView: ImageView
+    ) {
+        require(dataset.categories.size >= targetViews.size) {
+            "There must be enough categories in the dataset to display one category on each target" +
+                    " ImageView"
+        }
```
NicolasVial, line 53: Should we use a string ressource for this? Or because it is a require statement, it shouldn't be a string ressource ?

NicolasVial (commented):
Here is a first review even if some tests refactoring are missing. Nothing to say, everything is clear and looks correct to me. Good job :)

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningFragment.kt
Nitwix: Yes, thanks for reminding me, I'll add it now

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/presenter/LearningPresenter.kt
Nitwix: It will result in a crash, and the message is really a debug message, so I don't think it needs to be a resource. Ideally it should be handled with a feedback to the user saying the dataset is invalid or sth like that.

Nitwix:
The following tests fail. I will relaunch once the PR has been approved in order to not overload the CI.

`com.github.HumanLearning2021.HumanLearningApp.model.DummyDatabaseServiceTest > removeCategoryAlsoRemovesFromDatasets FAILED
    java.util.ConcurrentModificationException at DummyDatabaseServiceTest.kt:170

com.github.HumanLearning2021.HumanLearningApp.model.DummyDatabaseServiceTest > removeCategoryWorks FAILED
    java.util.ConcurrentModificationException at DummyDatabaseServiceTest.kt:162

com.github.HumanLearning2021.HumanLearningApp.model.DummyDatabaseManagementTest > removeCategoryWorks FAILED
    java.util.ConcurrentModificationException at DummyDatabaseManagementTest.kt:134

com.github.HumanLearning2021.HumanLearningApp.model.DummyDatabaseManagementTest > getDatasetsWorks FAILED
    java.lang.AssertionError at DummyDatabaseManagementTest.kt:228`

NicolasVial (approved):
Looks good to me !

wattsecs (approved):
Code is very readable, good job

Nitwix:
The `ConcurrentModificationException`s came from the dummy database service. The cause is that we modify a collection while it is being iterated on. This will take some time to fix, and I don't have the courage to start this now. So I will merge this in next sprint.

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningTest.kt
```diff
@@ -51,29 +51,114 @@ class LearningTest {
...
     @Test
     fun dragImageOnCorrectCategory() {
+        launchFragment(firstDatasetId)
         val imToSort = getImageToSort()
         val startDescr = imToSort.contentDescription
         val NUMBER_OF_ATTEMPTS = 50
         var foundImageChange = false
         // tries drag and drop NUMBER_OF_ATTEMPTS times until there is a change in the image to sort
-        // (change in description)
+        // (change in category)
```
Nitwix: TODO: change this back

PR #156, opened 2021-05-10
Refactor PictureRepository

wattsecs:
I kept the `CachePictureRepository` (now `CachePictureStorage`) because I believe that it prevents mistakes in accessing the wrong cache directory.


bbjubjub2494 (changes requested):
If I understand correctly, the point of the PR is to rename both classes. I don't know if this has been discussed somewhere else, but I don't understand how this is storage rather than a repository. It's basically CRUD, (although both `savePicture` should probably be rethought) which is the essence of the "repository pattern" as I understand it

wattsecs:
> 
> 
> If I understand correctly, the point of the PR is to rename both classes. I don't know if this has been discussed somewhere else, but I don't understand how this is storage rather than a repository. It's basically CRUD, (although both `savePicture` should probably be rethought) which is the essence of the "repository pattern" as I understand it

Yes and to add documentation, I did this refactoring now because I was feeling like it basically.
The name change is following the code review by the assistants that, if I understood correctly, is mainly because this does not follow the Repository pattern and should hence not be named as such.

bbjubjub2494:
> > If I understand correctly, the point of the PR is to rename both classes. I don't know if this has been discussed somewhere else, but I don't understand how this is storage rather than a repository. It's basically CRUD, (although both `savePicture` should probably be rethought) which is the essence of the "repository pattern" as I understand it
> 
> Yes and to add documentation, I did this refactoring now because I was feeling like it basically.
> The name change is following the code review by the assistants that, if I understood correctly, is mainly because this does not follow the Repository pattern and should hence not be named as such.

Oh, I think I misunderstood. Then here are my updated thoughts:

1. "Storage" is unsuitable as a noun substantive in a class name because it's uncountable. I think "cache" is the next best thing. (see #2) Or we could CRUDify it but that takes more time.
2. The reviewers  did not understand the purpose of `CachePictureRepository`; they thought it was unfinished. That might be an indication that we should replace it with a factory function. That can help with just calling it `ImageCache` or something.

wattsecs:
Converted back into draft because it probably won't be finished this week

wattsecs:
Closed for now because it is no longer relevant. If a refactor is done at a later point it will be reopened.

PR #158, opened 2021-05-11
Feature/145 user separation UIs

### LINKED ISSUE #145: Adapt UI to allow for user separation between learner and admin

NicolasVial:
For the moment :

- When not logged in or logged in, not as admin : have the "start learning !" button and the bottom navigation tab with the Learning button.
- When logged in, as admin : have the "start learning !" button and the bottom navigation tab with the Learning button and the  Dataset Editing button.

I am open to any modification that would make it better for the user. 

Comment : 

- Need Marouane PR to be merged before this one to be merged


app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/MainActivity.kt
```diff
@@ -41,11 +54,28 @@ class MainActivity : AppCompatActivity() {
...
+        val bottomNav = findViewById<BottomNavigationView>(R.id.bottom_nav)
+        bottomNav.setupWithNavController(navController)
+        val goToLearningButton = bottomNav?.menu?.get(0)
+        val goToDsEditingButton = bottomNav?.menu?.get(1)
+
+        navController.addOnDestinationChangedListener { _, _, _ ->
+            lifecycleScope.launch {
+                val user = presenter.currentUser
+                val isAdmin = user?.let { dbService.checkIsAdmin(it) }
+                if (user == null || !isAdmin!!) {
```
wattsecs: `user == null` is not necessary since `isAdmin` would be null in this case. I suggest doing something like `isAdmin?.let { if (it) ....} ?: ...`

wattsecs (changes requested):

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/MainActivity.kt
NicolasVial: Oh yes you're right Thank you ! Forgot to modify that. (Btw don't review all the "file changes" since a lot of those changes are from Marouane's PR.)

wattsecs (approved):
All good

Jaakik (approved):

PR #160, opened 2021-05-11
Integrate caching

wattsecs:
Integrates caching into the production dependency injection


app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/offline/CachedDatabaseService.kt
```diff
@@ -48,12 +61,23 @@ class CachedDatabaseManagement internal constructor(
...
+    private suspend fun removeFromCache(pictureId: Id) {
         cachedPictures.remove(pictureId)
         try {
             cache.deletePicture(pictureId)
         } catch (e: DatabaseService.NotFoundException) {
             // Do nothing since it means that the cache already removed it itself
         }
     }
+
+    private suspend fun getRepresentativePictureFromCache(pictureId: Id): CategorizedPicture? {
```
bbjubjub2494, line 68: I don't really understand what this function does and why it needs to be this convoluted. It might be good to briefly explain what problem it's supposed to solve.

bbjubjub2494 (changes requested):
Well executed overall, good handling of blocking IO. Just a couple things to say

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/offline/CachedDatabaseService.kt
```diff
@@ -2,23 +2,29 @@ package com.github.HumanLearning2021.HumanLearningApp.offline
 
 import com.github.HumanLearning2021.HumanLearningApp.model.*
 
-class CachedDatabaseManagement internal constructor(
-    private val db: DatabaseManagement, private val cache: PictureRepository
-) : DatabaseManagement by db {
+/**
+ * Decorator for a DatabaseManagement which adds caching for the pictures
```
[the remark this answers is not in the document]
wattsecs: Indeed, thank you

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/offline/CachedDatabaseService.kt
wattsecs: This function is called if the representative picture has already been cached for this category (line 40) and then tries to get it from cache or reloads it if it is no longer in there.
This makes me think that having a list which keeps track of representative pictures is a bad idea since those could change without it being updated. I will remove this list and always retrieve the picture id of the representative picture from firestore instead.

martin-lenweiter (approved):
Looks good. Readable code, and brings us one step closer to making all the requirements

bbjubjub2494 (approved):
LGTM

PR #166, opened 2021-05-11
Feature#125/UserAdminAccess

Jaakik:
I have Implemented Changes to the Backend to account for User vs Learner Mode 

TODO : 

- Valdiate Backend changes with the Team

- Add checkBox in Login 

- Use DemoDB instead of Dummy for login


app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/DatabaseService.kt
```diff
@@ -184,6 +184,12 @@ interface DatabaseService {
 
     suspend fun updateUser(firebaseUser: FirebaseUser): User
 
+    suspend fun setAdminAccess(firebaseUser: FirebaseUser, adminAccess: Boolean): User
+
+    suspend fun checkIsAdmin(user : User): Boolean
```
bbjubjub2494, line 189: is this method redundant? should it refresh into the database?

Jaakik: Yes it should, should directly implement this in UpdateUser with argument AdminAccess ?

bbjubjub2494: I can see that all the implementations right now just check the boolean, but if this is more futureproof go ahead.

>should [I] directly implement this in UpdateUser with argument AdminAccess ?

`setAdminAccess` is fine, I was asking about `checkIsAdmin`

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/firestore/FirestoreDatabaseService.kt
```diff
@@ -309,6 +311,22 @@ class FirestoreDatabaseService internal constructor(
         return documentRef.get().await().toObject(UserSchema::class.java)!!.toPublic()
     }
 
+    //TODO Optimize transaction to db
+    override suspend fun setAdminAccess(firebaseUser: FirebaseUser ,adminAccess : Boolean): User {
```
wattsecs: Using the [update method](https://firebase.google.com/docs/reference/js/firebase.firestore.DocumentReference#update) would improve this

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/firestore/FirestoreDatabaseService.kt
```diff
@@ -309,6 +311,22 @@ class FirestoreDatabaseService internal constructor(
...
+        val documentRef = users.document("$uid@$type")
+        val data = UserSchema().apply {
+            isAdmin = adminAccess
+        }
+        documentRef.set(data).await()
+        return documentRef.get().await().toObject(UserSchema::class.java)!!.toPublic()
+    }
+
+    override suspend fun checkIsAdmin(fireStoreUser: User): Boolean {
+        return fireStoreUser!!.isAdmin
```
wattsecs: I don't think the non-null assertion is necessary here?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/offline/OfflineDatabaseService.kt
```diff
@@ -248,12 +248,35 @@ class OfflineDatabaseService internal constructor(
                 firebaseUser.uid,
                 User.Type.FIREBASE,
                 firebaseUser.displayName,
-                firebaseUser.email
+                firebaseUser.email,
+                getUser(User.Type.FIREBASE, firebaseUser.uid)!!.isAdmin,
```
wattsecs: I suggest using the ?. operator and return the non-updated user if getUser() returns null in combination with logging that the operation failed

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/offline/OfflineDatabaseService.kt
```diff
@@ -248,12 +248,35 @@ class OfflineDatabaseService internal constructor(
...
+        userDao.update(
+            RoomUser(
+                firebaseUser.uid,
+                User.Type.FIREBASE,
+                firebaseUser.displayName,
+                firebaseUser.email,
+                adminAccess,
+            )
+        )
+        return fromUser(userDao.load(firebaseUser.uid, User.Type.FIREBASE)!!)
```
wattsecs: I suggest using the ?. operator and return the non-updated user if getUser() returns null in combination with logging that the operation failed

wattsecs (changes requested):
Please use the code format tool to avoid empty lines and misplaced white spaces around "," (see https://github.com/HumanLearning2021/HumanLearningApp/pull/153 in combination with the Save Action plugin)

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/firestore/FirestoreDatabaseService.kt
Jaakik: it is not , just by reflex , I ll remove it

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/offline/OfflineDatabaseService.kt
Jaakik: Thanks for suggesting that, I was looking for an alternative

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/firestore/FirestoreDatabaseService.kt
Jaakik: You mean by changing the signature of update method to be the same as setAdmin ? I thought about that but for further implementations I think there's going to be a fundamental difference between setAdmin and update as more checks will be happening to approve of user as admin

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/offline/OfflineDatabaseService.kt
Jaakik: for that isAdmin should be of type boolean?  ? ( second question mark is a literal question mark )

wattsecs: No, I mean the result of `userDao.load`. The objective would be to avoid crashes because of NullPointerErrors which could occur with the usage of `!!`

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/firestore/FirestoreDatabaseService.kt
wattsecs: No, I mean changing how the document is updated by using `documentRef.update(...` instead of `documentRef.set(...`

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/firestore/FirestoreDatabaseService.kt
```diff
@@ -309,6 +311,22 @@ class FirestoreDatabaseService internal constructor(
...
+        val type = User.Type.FIREBASE
+        val documentRef = users.document("$uid@$type")
+        val data = UserSchema().apply {
+            isAdmin = adminAccess
+        }
+        documentRef.set(data).await()
+        return documentRef.get().await().toObject(UserSchema::class.java)!!.toPublic()
+    }
+
+    override suspend fun checkIsAdmin(fireStoreUser: User): Boolean {
```
wattsecs, line 360: Would this method not be more useful if it took a user id as an argument?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/offline/OfflineDatabaseService.kt
wattsecs: Please adress this comment

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/offline/OfflineDatabaseService.kt
```diff
@@ -248,12 +248,22 @@ class OfflineDatabaseService internal constructor(
...
-                firebaseUser.email
+                firebaseUser.email,
+                getUser(User.Type.FIREBASE, firebaseUser.uid)!!.isAdmin,
             )
         )
         return fromUser(userDao.load(firebaseUser.uid, User.Type.FIREBASE)!!)
     }
 
+    override suspend fun setAdminAccess(firebaseUser: FirebaseUser, adminAccess: Boolean): User {
+        throw Exception("No Admin In Offline Mode Error")
```
wattsecs: Creating a custom exception or at least using `IllegalStateException` would be more descriptive

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/offline/OfflineDatabaseService.kt
```diff
@@ -248,12 +248,22 @@ class OfflineDatabaseService internal constructor(
...
+    }
+
+    override suspend fun checkIsAdmin(user: User): Boolean {
+        throw Exception("No Admin In Offline Mode Error")
```
wattsecs: Creating a custom exception or at least using `IllegalStateException` would be more descriptive

wattsecs (changes requested):
The login UI from a user standpoint is a bit unclear (the funcitonality of the admin check box is not very clear) but this can be improved later

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/firestore/FirestoreDatabaseService.kt
```diff
@@ -309,6 +311,22 @@ class FirestoreDatabaseService internal constructor(
         return documentRef.get().await().toObject(UserSchema::class.java)!!.toPublic()
     }
 
+    //TODO Optimize transaction to db
```
wattsecs: What about this?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/firestore/FirestoreDatabaseService.kt
Jaakik: For me No , but we have to ask @NicolasVial who uses the backend for the UI separation flow

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/firestore/FirestoreDatabaseService.kt
Jaakik: Yes It's adressed I only update the admin Instead of updating all the user variables

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/firestore/FirestoreDatabaseServiceTest.kt
```diff
@@ -312,4 +321,15 @@ class FirestoreDatabaseServiceTest {
     fun test_getRepresentativePicture() = runBlocking {
         assertThat(demoInterface.getRepresentativePicture(appleCategoryId), not(equalTo(null)))
     }
+
+    @Test
+    fun test_setAdminAccess() {
+        runBlocking {
+            val firebaseUser = Firebase.auth.signInAnonymously().await().user!!
+            presenter.onSuccessfulLogin(false)
+            val fireStoreUser = scratchInterface.setAdminAccess(firebaseUser, true)
```
bbjubjub2494:
```suggestion
            val user = scratchInterface.setAdminAccess(firebaseUser, true)
```
- having two fire[blah]User is confusing
- The `User` interface is a thing
- The `FirestoreUser` is going away

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/offline/OfflineDatabaseService.kt
```diff
@@ -234,17 +234,18 @@ class OfflineDatabaseService internal constructor(
...
-        userDao.update(
-            RoomUser(
-                firebaseUser.uid,
-                User.Type.FIREBASE,
-                firebaseUser.displayName,
-                firebaseUser.email
-            )
-        )
-        return fromUser(userDao.load(firebaseUser.uid, User.Type.FIREBASE)!!)
+        throw IllegalStateException("No Users Downloaded yet")
```
bbjubjub2494, line 237: If we intend to catch it (?) it should be a specific exception class, not a generic one.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/GoogleSignInWidget.kt
```diff
@@ -100,5 +109,6 @@ class GoogleSignInWidget : Fragment() {
 
     companion object {
         val RC_SIGN_IN: Int = "firebase auth".hashCode()
+        var isAdmin = false
```
bbjubjub2494: Shouldn't this be in the instance?

bbjubjub2494 (changes requested):

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/offline/OfflineDatabaseService.kt
Jaakik: We do not intend to catch it for now I believe, it should be implemented in a later PR by @jonaskon  ?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/GoogleSignInWidget.kt
Jaakik: I just found it easier to pass around in this fashion to represent the global state of the loginUI

bbjubjub2494: Why can it not be an instance variable on the Fragment? I don't understand your justification.

EDIT: I get why you did this. it's still a bad pattern. Do I have your permission to patch it?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/DatabaseService.kt
bbjubjub2494: It is not clear at all from the perspective of the caller what the difference between `user.isAdmin` and `checkIsAdmin(user)` is. Please explain

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/firestore/FirestoreDatabaseService.kt
NicolasVial: I think it is easy like this, we can get the user with the presenter  using authPresenter.currentUser.

wattsecs (approved):
My requested changes were implemented. Ok for me

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/GoogleSignInWidget.kt
Jaakik: Yes Please

bbjubjub2494: done, thank you

bbjubjub2494 (approved):

Issue #177, opened 2021-05-14
Partially refactor data repository back-end


bbjubjub2494:
Per #151, as much as possible there should be POKOs and generic Repository Pattern as much as possible.

PR #69, opened 2021-04-06
Feature/43 firestore database management

wattsecs:

PR #179, opened 2021-05-14
Remove unused ConcurrentDeleteDatabaseManagement

bbjubjub2494:
For #177 I'm going to submit multiple PRs in increasing order of invasiveness

This PR simply removes some dead code.


Nitwix (approved):
Always good to do some cleanup.
But what was the original purpose of this class?

bbjubjub2494:
> Always good to do some cleanup.
> But what was the original purpose of this class?

This originally comes from a commit from Jonas titled "removeCategoryFromDataset now returns a dataset even if the underlying databaseService throws an exception" in #69. I preserved and enhanced the functionality when refactoring, but then we never actually used it. It is also more difficult to implement with the class hierarchy refactor, that's what prompted me to toss it.

NicolasVial (approved):

PR #182, opened 2021-05-14
[housekeeping] Bump dependencies

bbjubjub2494:
- Kotlin 1.5 breaks jacoco reports, hence it has been skipped
- Android SDK 4.2 and/or gradle 6.7 break some stuff, hence they are not there

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -45,7 +45,7 @@ android {
     buildFeatures {
         viewBinding true
     }
-    
+
     viewBinding {
         enabled = true
     }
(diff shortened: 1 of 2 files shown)


wattsecs (commented):
I would also suggest bumping the gradle version to 4.2.1 (line 11 in build.gradle)

bbjubjub2494:
> I would also suggest bumping the gradle version to 4.2.1 (line 11 in build.gradle)

very good suggestion. (unless you're like me and you haven't updated Studio, but that's on me)

wattsecs (approved):

NicolasVial (approved):

PR #183, opened 2021-05-14
[bugfix] fix a bug occuring when setting a representative picture

wattsecs:
Fix a bug where setting a representative picture from a categorized picture already present in the dataset would lead to said picture being deleted from storage. This caused a crash because the app would try to load an image from a non-existant url.


Nitwix (approved):
It seems to fix the crashes, indeed. Good job!

NicolasVial (approved):
Excellent ! Thank you for the fix

PR #188, opened 2021-05-17
Refactor/176 : various bugifxes and make some UIs clearer for users

### LINKED ISSUE #168: The title displayed on the DisplayImageSetFragment is not user friendly
sth like nav_graph_label_display_image_s...
instead of the name of the fragment

### LINKED ISSUE #176: Various bugfix and make some UIs clearer

NicolasVial:
Bugfixes : 

- Title displayed are now user friendly.
- Views are not moved anymore when opening the keyboard (it was moving the views above the keyboard which made the views to walk on each other)
- Cannot modify the name of the already existing categories in CategoriesEditing fragment.

UIs upgrades : 

- CategoriesEditing : It's now here that we can modify the dataset name. There is a "info" menu that explains some stuff (wip) to the user. When creating a datastet, already have 3 empty categories for the understanding of the user.
- DisplayImageset : Added an info icon that creates a popup when clicked that explains how to delete images + how to set as representative picture.

Comments : 

- The texts in the Info popups need to be modified, open to any idea.


app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/MetadataEditingFragment.kt
```diff
@@ -63,7 +64,12 @@ class CategoriesEditingFragment : Fragment() {
                     new = true
                 }
                 if (!new) {
+                    /**
+                     * If the dataset is not new, we add the already existing categories
+                     * and the name of the dataset to the layout and set them as non clickable.
+                     */
                     dataset = dBManagement.getDatasetById(datasetId!!)!!
```
Jaakik: Maybe the the not-null assertion operator is not necessary here as you already perform the check above

Jaakik (approved):
Looks good :))

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/MetadataEditingFragment.kt
NicolasVial: I agree, your idea is correct but it is necessary :( I changed the !! to a ?. for the datasetId because it's cleaner I think but for the call to getDSById, I think we can consider that we always receive an existing DSId without problem. Correct me if I am wrong.

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/DisplayDatasetActivityTest.kt
```diff
@@ -136,29 +134,9 @@ class DisplayDatasetActivityTest {
...
-    fun modifyingDatasetNameWorks() {
-        launchFragment()
-        val newName = "new dataset name"
-        runBlocking {
-            waitFor(1) // increase if needed
-            onView(withId(R.id.display_dataset_name)).perform(clearText(), typeText("$newName\n"))
-            onView(withId(R.id.display_dataset_name)).check(matches(withText(containsString(newName))))
-
-            // need to get again because dataset is immutable and editing the name creates a new
-            // Dataset object in the database
```
martin-lenweiter, line 149: Subjective, but I would find it more intuitive if modifying the dataset name wasn't in the modify categories menu item. We could either add a separate menu item, or do it in datasetoverview maybe?

app/src/main/res/values/strings.xml
```diff
@@ -45,8 +47,13 @@
     <string name="setAsRepresentativePicture">Set as representative picture</string>
     <string name="deletingDataset">Deleting dataset</string>
     <string name="deleteConfirmationMessage">Are you sure you want to delete this dataset?</string>
+    <string name="categoriesEditingInfo">- A category is a word representing a type of object (Ex: Fork) -\n\n
+        - Every picture belong to a category of the dataset -</string>
```
martin-lenweiter:
```suggestion
        - Every picture belongs to a category of the dataset -</string>
```

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/DisplayImageSetFragment.kt
```diff
@@ -245,4 +247,25 @@ class DisplayImageSetFragment : Fragment() {
...
         })
     }
+
+    override fun onCreateOptionsMenu(menu: Menu, inflater: MenuInflater) {
+        inflater.inflate(R.menu.display_imageset_info_menu, menu)
+        super.onCreateOptionsMenu(menu, inflater)
+    }
+
+    override fun onOptionsItemSelected(item: MenuItem): Boolean {
+        return when (item.itemId) {
```
martin-lenweiter, line 257: Great addition.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/MetadataEditingFragment.kt
```diff
@@ -25,13 +25,13 @@ import kotlinx.coroutines.launch
 import javax.inject.Inject
 
 @AndroidEntryPoint
-class CategoriesEditingFragment : Fragment() {
+class MetadataEditingFragment : Fragment() {
```
martin-lenweiter, line 28: I like this renaming, actually that would discard my comment above. But then let's also change the menu item's name accordingly.

martin-lenweiter (approved):
All the bugs seem to have been fixed and some nice clarifications in the UI. Only a small request, but can be merged as is.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/MetadataEditingFragment.kt
NicolasVial: Forgot this one you're right ! I will update that :)

PR #180, opened 2021-05-14
Refactor/177 Flatten POKO hierarchy

bbjubjub2494:
For #177 I'm going to submit multiple PRs in increasing order of invasiveness.

This PR flattens the POKO hierarchy as recommended in the code review.

Note: It adds a ton of deprecation warnings. That's so as to minimize the amount of lines touched and potential conflicts. At some point where we have no PRs outstanding, we can auto-refactor those out.

PR #192, opened 2021-05-18
Refactor DummyDatabaseService using the Repository Pattern

### LINKED ISSUE #177: Partially refactor data repository back-end

bbjubjub2494:
Depends on #180 

Close #177 

It should be possible to write a Repository for a Firestore collection, a couple Repository adapters, and ultimately dismantle DatabaseService.

### DIFF
--- a/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/InMemoryRepository.kt
+++ b/app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/InMemoryRepository.kt
@@ -1,0 +1,58 @@
+package com.github.HumanLearning2021.HumanLearningApp.model
+
+import android.util.Base64
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import kotlin.random.Random
+
+/**
+ * In-memory, impermanent key-value storage.
+ */
+class InMemoryRepository<V>(
+    private val random: Random = Random,
+) : Repository<String, V> {
+
+    private inner class Context {
+        var data = mutableMapOf<String, V>()
+    }
+
+    private val mutex = Mutex()
+    private val ctx = Context()
+    private suspend fun <R> withContext(action: Context.() -> R): R = mutex.withLock {
+        ctx.action()
+    }
+
+    suspend fun createWith(fkv: (String) -> V) = withContext {
+        var k: String
+        do {
+            k = Base64.encodeToString(random.nextBytes(15), Base64.URL_SAFE)
+        } while (k in data)
+        data[k] = fkv(k)
+        k
+    }
+
+    override suspend fun updateWith(k: String, f: (V?) -> V?) = withContext {
+        f(data[k])?.let { data[k] = it }
+        Unit
+    }
+
+    override suspend fun getById(k: String) = withContext {
+        data[k]
+    }
+
+    override suspend fun delete(k: String): Boolean = withContext { data.remove(k) != null }
+
+    override suspend fun getIds() = withContext { data.keys }
+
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/InMemoryRepository.kt
```diff
@@ -0,0 +1,58 @@
+package com.github.HumanLearning2021.HumanLearningApp.model
+
+import android.util.Base64
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import kotlin.random.Random
+
+/** In-memory, impermanent key-value storage.
```
wattsecs:
```suggestion
/** 
/* In-memory, impermanent key-value storage.
```

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/InMemoryRepository.kt
```diff
@@ -0,0 +1,58 @@
...
+    private val mutex = Mutex()
+    private val ctx = Context()
+    private suspend fun <R> withContext(action: Context.() -> R): R = mutex.withLock {
+        ctx.action()
+    }
+
+    suspend fun createWith(fkv: (String) -> V) = withContext {
+        var k: String
+        do {
+            k = Base64.encodeToString(random.nextBytes(15), Base64.URL_SAFE)!!
```
wattsecs:
```suggestion
            k = Base64.encodeToString(random.nextBytes(15), Base64.URL_SAFE)
```

wattsecs (changes requested):
Small typos

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/InMemoryRepository.kt
bbjubjub2494: The compiler complains about nullability inference. We have to explicitly assert that `encodeToString` does not return null.

wattsecs: I don't understand because according to the [doc](https://developer.android.com/reference/android/util/Base64#encodeToString(byte[],%20int)) the return value is already non-null and compiling on my machine did not throw any warnings ?

bbjubjub2494: > I don't understand because according to the [doc](https://developer.android.com/reference/android/util/Base64#encodeToString(byte%5B%5D,%20int)) the return value is already non-null and compiling on my machine did not throw any warnings ?

It does for me too in the latest commit, my bad.

Jaakik (approved):

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/model/InMemoryRepository.kt
```diff
@@ -0,0 +1,58 @@
...
+
+    private inner class Context {
+        var data = mutableMapOf<String, V>()
+    }
+
+    private val mutex = Mutex()
+    private val ctx = Context()
+    private suspend fun <R> withContext(action: Context.() -> R): R = mutex.withLock {
+        ctx.action()
+    }
```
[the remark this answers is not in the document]
bbjubjub2494: I'm not sure if it needed to be concurrency safe, but now it is. Consider that a bonus feature.

wattsecs (approved):

PR #202, opened 2021-05-23
Feature#198/login persistance

Jaakik:


app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/MainActivity.kt
```diff
@@ -76,7 +82,11 @@ class MainActivity : AppCompatActivity() {
...
                 val user = authPresenter.currentUser
-                val isAdmin = user?.isAdmin ?: false
+                var isAdmin = user?.isAdmin ?: false
+                isAdmin.let {
+                    goToDsEditingButton?.isVisible = it
+                }
+                isAdmin = prefs!!.getBoolean("hasLogin", false)
                 isAdmin.let {
                     goToDsEditingButton?.isVisible = it
                 }
```
NicolasVial: I don't understand why you have 2 times the isAdmin.let{...}

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/MainActivity.kt
```diff
@@ -33,6 +34,9 @@ class MainActivity : AppCompatActivity() {
     @Inject
     lateinit var authPresenter: AuthenticationPresenter
 
+    var prefs: SharedPreferences? = null
+
```
NicolasVial: use lateinit var ?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/GoogleSignInWidget.kt
```diff
@@ -14,17 +18,22 @@ import androidx.fragment.app.Fragment
...
 
 @AndroidEntryPoint
 class GoogleSignInWidget : Fragment() {
+
     @Inject
     lateinit var presenter: AuthenticationPresenter
 
     var isAdmin = false
+    var prefs: SharedPreferences? = null
+    var editor: SharedPreferences.Editor? = null
```
NicolasVial: lateinit vars? and editor could be private.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/GoogleSignInWidget.kt
```diff
@@ -36,13 +45,34 @@ class GoogleSignInWidget : Fragment() {
...
             onLoginButtonPress()
         }
 
         view.findViewById<Button>(R.id.checkBox).setOnClickListener {
             isAdmin = view.findViewById<CheckBox>(R.id.checkBox).isChecked
         }
+        view.findViewById<Button>(R.id.singOutButton).setOnClickListener {
+            onSignOutPress()
+
+        }
```
NicolasVial: useless space

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/GoogleSignInWidget.kt
```diff
@@ -36,13 +45,34 @@ class GoogleSignInWidget : Fragment() {
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
+        prefs = activity?.getSharedPreferences("LOGIN", MODE_PRIVATE)
+        editor = prefs!!.edit()
```
NicolasVial: you have a warning... maybe this can help? [https://stackoverflow.com/questions/42092753/sharedpreferences-edit-without-a-corresponding-commit-or-apply-call](url) .

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/GoogleSignInWidget.kt
```diff
@@ -1,8 +1,12 @@
 package com.github.HumanLearning2021.HumanLearningApp.view
 
 import android.app.Activity
+import android.content.Context
+import android.content.Context.MODE_PRIVATE
 import android.content.Intent
+import android.content.SharedPreferences
 import android.os.Bundle
+import android.preference.PreferenceManager
```
NicolasVial: I think your imports could be optimized

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/GoogleSignInWidget.kt
```diff
@@ -94,16 +137,50 @@ class GoogleSignInWidget : Fragment() {
...
+                prefs!!.getString("email", ""),
+                prefs!!.getString("uid", "")!!,
+                User.Type.FIREBASE,
+                prefs!!.getBoolean("isAdmin", false)
+            )
+            presenter.currentUser = user
+            setSingOutUi()
+
+        } else {
+            val user = presenter.currentUser
```
NicolasVial: you have a shadow with you user var/val

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/presenter/AuthenticationPresenter.kt
```diff
@@ -48,7 +58,7 @@ class AuthenticationPresenter @Inject constructor(
     /**
      * Return information about the currently signed-in user
      */
-    val currentUser: User?
+    var currentUser: User? = null
```
NicolasVial: why this modification ?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/presenter/AuthenticationPresenter.kt
```diff
@@ -26,6 +30,11 @@ class AuthenticationPresenter @Inject constructor(
             .build()
     }
 
+    suspend fun signOut() {
+        Firebase.auth.signOut()
```
NicolasVial: Looks like it is  a redundent suspend identifier.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/presenter/AuthenticationPresenter.kt
```diff
@@ -1,6 +1,9 @@
 package com.github.HumanLearning2021.HumanLearningApp.presenter
 
+import android.content.Context
+import android.content.Context.MODE_PRIVATE
 import android.content.Intent
+import android.content.SharedPreferences
```
NicolasVial: you can optimize import I think?

NicolasVial (changes requested):
Here are already some comments. And I also have a question : When we login, the bottom nav. is not updated which mean that if we are logged in and we log out, we can "hack" the app and still go to the dataset editing interfaces. Do you think it is possible to modify bottom nav just after login/logout ?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/MainActivity.kt
Jaakik: I have noticed and addressed that

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/GoogleSignInWidget.kt
```diff
@@ -94,16 +137,50 @@ class GoogleSignInWidget : Fragment() {
         }
     }
 
+    private fun setSingInUi() {
+        view?.findViewById<Button>(R.id.singOutButton)?.visibility = View.GONE
+        view?.findViewById<Button>(R.id.checkBox)?.visibility = View.VISIBLE
+        view?.findViewById<Button>(R.id.loginButton)?.visibility = View.VISIBLE
+    }
+
+    private fun setSingOutUi() {
```
NicolasVial: I forgot this one haha,  I am sure that you're a good singer but I think sign is better here. ;)

NicolasVial (commented):
forgot a typo

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/GoogleSignInWidgetTest.kt
```diff
@@ -63,10 +83,52 @@ class GoogleSignInWidgetTest {
...
+        }
+        onView(withId(R.id.singOutButton)).check(matches((isDisplayed())))
+
+    }
+
+
+    @Test
+    fun signOutUserSuccess() {
+        runBlocking {
+            launchFragment()
```
martin-lenweiter: Usually calling launchFragment in setup is enough.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/presenter/AuthenticationPresenter.kt
martin-lenweiter: If you're using the saveAction plugin it should do it automatically I think.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/presenter/AuthenticationPresenter.kt
martin-lenweiter: yep doesn't look necessary

Jaakik: I have to do like this because now when signing out I have  set   _currentUser to  null

bbjubjub2494: that makes the `var` duplicative of `_currentUser`, which is asking for trouble at this point.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/GoogleSignInWidget.kt
```diff
@@ -94,16 +137,50 @@ class GoogleSignInWidget : Fragment() {
...
+            view?.findViewById<TextView>(R.id.loginStatus)?.text =
+                prefs?.getString("name", "Not logged in!")
+            user = User(
+                prefs!!.getString("name", ""),
+                prefs!!.getString("email", ""),
+                prefs!!.getString("uid", "")!!,
+                User.Type.FIREBASE,
+                prefs!!.getBoolean("isAdmin", false)
+            )
+            presenter.currentUser = user
```
bbjubjub2494: It's not the role of the view to tell the presenter what the current user is. This is incorrectly designed.

bbjubjub2494 (changes requested):
I don't think it's an acceptable solution to store (duplicatively) the user's profile in the preference and then fake a log-in in the interface. The model (or the firebase library) should be authoritative on user status, especially given that authentication happens in the back-end.

The correct solution in my view, would be to make sure that the log-in state is accurate in the presenter.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/GoogleSignInWidget.kt
Jaakik: Yes Agreed also this piece of code is unnecessary after using the shared preferences I will remove it

bbjubjub2494:
> I don't think it's an acceptable solution to store (duplicatively) the user's profile in the preference and then fake a log-in in the interface. The model (or the firebase library) should be authoritative on user status, especially given that authentication happens in the back-end.
> 
> The correct solution in my view, would be to make sure that the log-in state is accurate in the presenter.

Discussed in private, this is actually a relatively common practice.

bbjubjub2494 (approved):

app/src/main/res/layout/activity_main.xml
```diff
@@ -26,7 +26,10 @@
             android:layout_height="match_parent"
             android:layout_above="@id/bottom_nav"
             android:layout_below="@+id/toolbar"
+            android:layout_alignParentEnd="true"
             android:layout_marginTop="0dp"
+            android:layout_marginEnd="4dp"
+            android:layout_marginBottom="0dp"
```
NicolasVial, line 32: this is useless nah? well anyway it's a detail.

NicolasVial (approved):
some details are not perfect but mergable

app/src/main/res/layout/activity_main.xml
Jaakik: I switched the position of certain buttons because they were hidden at times

Jaakik:
> some details are not perfect but mergable

Yes, they will be taken in the next sprint, mainly the signOut and datasetediting button, and rethink the login persistance

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/GoogleSignInWidgetTest.kt
Jaakik: Done :)

PR #204, opened 2021-05-24
Finish making learning UI clearer

### LINKED ISSUE #141: Finish making the learning UI clearer (see PR#140)
see : https://github.com/HumanLearning2021/HumanLearningApp/pull/140

The land, sw600dp, and sw600dp-land layouts in fragment_learning also need to be adapted

Related: desynchronization between view and model causes inconsistent feedback. (to test)

Nitwix:
closes #141 
This PR adds dynamic visual feedback for the learning fragment.
The element with which to interact blinks in blue, and upon correct or incorrect sorting, the image blinks rapidly in green or red.
Warning to people that are epileptic, there's quite a lot of blinking going on.


NicolasVial (commented):
Here is my opinion after that of Jonas: For me as regards the image to classify it is very clear. On the other hand, I would find it clearer and less aggressive for the eyes to simply have the correct/incorrect image surrounded by green or red (not flashing).
From a code point of view it looks good to me.

wattsecs (changes requested):
See https://github.com/HumanLearning2021/HumanLearningApp/pull/204#issuecomment-849412364

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningFragment.kt
```diff
@@ -94,18 +98,39 @@ class LearningFragment : Fragment() {
...
+
+        visualFeedback = with(binding) {
+            LearningVisualFeedback(
+                lifecycleScope,
+                baseColor = getColor(requireContext(), R.color.white),
+                neutralColor = getColor(requireContext(), R.color.blue),
+                positiveColor = getColor(requireContext(), R.color.light_green),
+                negativeColor = getColor(requireContext(), R.color.red),
+                // TODO remove !! once enclosing CardViews put in all layouts
+                sourceCardView = learningToSortCv!!,
```
NicolasVial, line 121: I don't perfectly understand this TODO. I mean what did you want to do to remove this todo?

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningFragment.kt
```diff
@@ -94,18 +98,39 @@ class LearningFragment : Fragment() {
...
+                    learningCat1Cv!!,
+                    learningCat2Cv!!
+                )
+            )
+        }.also { it.sourceCardViewShouldBlink(true) }
+
         requireActivity().onBackPressedDispatcher.addCallback(callback)
     }
 
+
```
NicolasVial, line 133: In the adaptDisplayToNumberOfCategories KDoc you have the wrong @param I think. (Sorry it's not from this PR but well..)

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningFragment.kt
```diff
@@ -216,11 +245,13 @@ class LearningFragment : Fragment() {
      */
     private fun dropCallback(event: DragEvent, v: View): Boolean {
```
NicolasVial: looks llike event is not used in this method

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningVisualFeedback.kt
```diff
@@ -0,0 +1,125 @@
...
+                blink(cv, baseColor, blinkColor, SHORT_BLINK_DURATION_MS)
+            }
+        }
+    }
+
+    companion object {
+        private val BLINK_DURATION_MS: Long = 800
+        private val SHORT_BLINK_DURATION_MS: Long = 150
+        private val NB_BLINKS_FOR_SORTING_FEEDBACK = 4
+    }
```
NicolasVial: use const ?

NicolasVial (changes requested):
I think it's good now ! The learning part looks really great now. some minor comments and 1 question about a todo :)

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningFragment.kt
Nitwix: I haven't added the cardview behind the image in all versions of the learning layout, so I need to put `!!` to use the cardviews. This is bad because the learning will crash (with a NullPointerException)  if you use it on a tablet for example. But I didn't have the time to adjust all the layouts...
I will add an issue so we don't forget to address this.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningFragment.kt
Nitwix: thanks x)

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningFragment.kt
Nitwix: yup, it's a relique from a long forgotten past. Thanks

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningVisualFeedback.kt
Nitwix: yup, thanks. (I'm still in the scala world where `val` is effectively a `const`)

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/view/learning/LearningFragment.kt
NicolasVial: oooh got it thanks

NicolasVial (approved):
good for me now :) Good job

wattsecs (approved):

PR #205, opened 2021-05-24
refactor databases

bbjubjub2494:
- test: exists in the Firebase Emulator, read-only, preloaded with specific data
- scratch: exists in the Firebase Emulator, read-write, contains whatever
- demo: exists on the Production cloud, not used in tests
- demo2: alias for demo, deprecated(?)

- [ ] update `firestore_state_windows`


bbjubjub2494:
superceded by #206, and I am no longer assigned to the issue

Issue #127, opened 2021-04-30
Merge "demo" and "demo2" databases


bbjubjub2494:

bbjubjub2494:
Some info:
- In the testing infrastructure as it is, we don't use Demo2, only Demo and on the FB emulator. In production, we don't really use Demo since it's frozen, so it would make sense to yeet demo from the cloud.
- We can yeet scratch from the Firestore but we need it on Cloud Storage currently.
- If we can decouple image storage backend and database backend we can further simplify DI.
- I won't have time to do the code this sprint.

PR #209, opened 2021-05-25
reorganize emulated databases #127

bbjubjub2494:
demo -> test
demo2 -> prod

I made prod writeable on the cloud. The demo will have to be performed on prod.

Merging scratch and test is not feasible unless we can reset the emulator every time because scratch data derails demo tests.


app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/hilt/modules.kt
```diff
@@ -36,23 +36,23 @@ annotation class DummyDatabase
```
wattsecs, line 39:
```suggestion
// Read only database used for testing purposes
```

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/hilt/modules.kt
```diff
@@ -36,23 +36,23 @@ annotation class DummyDatabase
 
 @Qualifier
 @Retention(AnnotationRetention.BINARY)
-annotation class DemoDatabase
+annotation class TestDatabase
 
 @Qualifier
 @Retention(AnnotationRetention.BINARY)
-annotation class CachedDemoDatabase
+annotation class CachedTestDatabase
```
wattsecs, line 50:
```suggestion
// Production database used by the app
```

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/hilt/modules.kt
```diff
@@ -36,23 +36,23 @@ annotation class DummyDatabase
...
 
 @Qualifier
 @Retention(AnnotationRetention.BINARY)
-annotation class Demo2Database
+annotation class ProdDatabase
```
wattsecs, line 55:
```suggestion
// Database used for testing only
```

wattsecs (changes requested):
I suggest adding some description to explain what each database is.
Also I looked into resetting the database between each test but there is no other way I found other than deleting it's entire content and then reimport the desired state. I think have demo and scratch is a better option.

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/model/UniqueDatabaseManagementTest.kt
```diff
@@ -46,18 +46,18 @@ class UniqueDatabaseManagementTest {
...
-        uDbMan.downloadDatabase("demo")
-        assertThat(uDbMan.getDownloadedDatabases(), hasItem("demo"))
+        uDbMan.downloadDatabase("test")
+        assertThat(uDbMan.getDownloadedDatabases(), hasItem("test"))
     }
 
     @Test
     fun offlineAndFirestoreDatabasesContainTheSameElements() = runBlocking {
-        val dbName = "demo"
+        val dbName = "test"
```
Nitwix: Maybe it would make sense to extract "test" into a `val` to ease future refactorings.

app/src/main/java/com/github/HumanLearning2021/HumanLearningApp/hilt/modules.kt
Nitwix: It would be even better in kdoc form no?
`/** */`

Nitwix (approved):
One minor comment, otherwise all good!

wattsecs (approved):

PR #233, opened 2021-06-01
task/224 give more complete information with de Info buttons.

### LINKED ISSUE #224: Rewrite Info text for DisplayImageSet and MetadataEditing fragments

NicolasVial:
This PR : 

- Now have more complete info texts about general infos + possible actions.
- i also fixed a bug in this PR because I did in 1 line of code (it's a dirty hack but I didn't find a way to do smth better). The bug was that the app crash if we long click on a picture in ImageSetFragment and then go back with the back arrow.
- un-ignore a test that is now fixed :)


wattsecs (changes requested):
~~I think this pr should also remove the long-press info because it requires too high of an Android version and is now obsolete~~
Please ignore this comment, I got confused... Sorry

wattsecs (approved):

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/MetadataEditingFragmentTest.kt
```diff
@@ -177,7 +175,7 @@ class MetadataEditingFragmentTest {
             navigateToCreateDatasetFragment()
             onView(withId(R.id.categories_editing_menu_info)).perform(click())
             waitFor(1) // increase if needed
```
martin-lenweiter, line 177: Would be nice to also test that the alert dialog is actually displayed. (I think I did test this in the learningSettingsFragment, so have a look there #235)

martin-lenweiter (changes requested):
Small improvement needed in one test, other than that all good

app/src/androidTest/java/com/github/HumanLearning2021/HumanLearningApp/view/dataset_editing/MetadataEditingFragmentTest.kt
NicolasVial: Oh yes ty you're right. i'll add it here and in ImageSetTest too :=)

martin-lenweiter (approved):

PR #234, opened 2021-06-01
Bug fix/privileges after signout

Jaakik:
* Bug is Fixed by moving directly to Homefragemnt after SignOut 
* Some Kdoc for signingWidget was added
* Removed Unecessary code to update the currentUser  in UpdateUI as it is not used anywhere else, the initial idea was to preserve consistency


Jaakik:
@mlenw I didn't deal with the BackPress as it seemed to work fine When testing in on my phone, please request any changes you think are necessary in this regard.

NicolasVial (approved):
Looks good to me but... are you sure about the KDOC? I mean... I think Jonas have a PR to add all the missing KDoc ^^' we don't want to have conflicts in the KDoc I think :)

app/src/main/res/navigation/nav_graph.xml
```diff
@@ -172,7 +172,11 @@
         android:id="@+id/googleSignInWidget"
         android:name="com.github.HumanLearning2021.HumanLearningApp.view.GoogleSignInWidget"
         android:label="@string/nav_graph_label_login"
-        tools:layout="@layout/fragment_google_sign_in" />
+        tools:layout="@layout/fragment_google_sign_in" >
+        <action
+            android:id="@+id/action_googleSignInWidget_to_homeFragment"
+            app:destination="@id/homeFragment" />
```
martin-lenweiter:
```suggestion
            app:destination="@id/homeFragment"
            app:popUpTo="@+id/homeFragment"
            app:popUpToInclusive="true"
            />
```

martin-lenweiter: When I press back it takes me back to login fragment. I think this should fix it.

martin-lenweiter (changes requested):
Back thing has an issue, and I agree with Nicolas about Kdoc. Need to be careful with merge conflicts. Other than that all good

Jaakik:
> Looks good to me but... are you sure about the KDOC? I mean... I think Jonas have a PR to add all the missing KDoc ^^' we don't want to have conflicts in the KDoc I think :)

Okay I will check that

martin-lenweiter (approved):

PR #240, opened 2021-06-02
ensure Android API level 24 compatibility

bbjubjub2494:


Nitwix:
In online mode, adding an existing picture works as expected on a phone with android 7.1.1.
But in offline mode, the app crashes when you try to display the picture that was added.
I am not sure though if this bug was introduced by this PR or if it existed before.
https://gist.github.com/Nitwix/1ffde82f634f4b089e07a04212deb09d

bbjubjub2494:
> In online mode, adding an existing picture works as expected on a phone with android 7.1.1.
> But in offline mode, the app crashes when you try to display the picture that was added.
> I am not sure though if this bug was introduced by this PR or if it existed before.
> https://gist.github.com/Nitwix/1ffde82f634f4b089e07a04212deb09d

Just tested, that's already in main afaict.

```
E/com.github.HumanLearning2021.HumanLearningApp.model.DefaultImageDisplayer: load failed
    class com.bumptech.glide.load.engine.GlideException: Failed to load resource
    There were 3 root causes:
    java.io.FileNotFoundException(/data/user/0/com.github.HumanLearning2021.HumanLearningApp/cache/04270b26-4728-495a-83a3-8fd83f3b83fe: open failed: ENOENT (No such file or directory))
    java.io.FileNotFoundException(open failed: ENOENT (No such file or directory))
    java.io.FileNotFoundException(open failed: ENOENT (No such file or directory))

Nitwix (approved):
Tested on a real device with android 7.1.1. Adding an existing picture works as expected. Good job!

martin-lenweiter (approved):
Seems to work fine

PR #242, opened 2021-06-03
Add credit for logo design.

Nitwix:
This PR adds two lines at the end of the README to credit the person who designed the logo for our app.


NicolasVial (approved):

wattsecs (approved):
Very good idea! If she has a website, Instagram or something we could also add a link to it

Nitwix:
> Very good idea! If she has a website, Instagram or something we could also add a link to it

It may be a good idea, but Nicolas tells me that she's actually a photographer, so it might be surprising if we link a page of hers and there is no graphic design on there. And anyway, a google search allows to find her info for people who would be interested.

PR #243, opened 2021-06-03
refactored string resource ids

Jaakik:
- Refactored String Resource ids as such : FragmentName_descriptionView
- Removed Unused Resources


martin-lenweiter (approved):

app/src/main/res/values/strings.xml
```diff
@@ -8,72 +8,53 @@
...
-    - Once you have finished everything you wanted to do, you can click on the "Save data" button to
-    save your modifications.</string>
-    <string name="displayImagesetInfo">
-        General information :\n\n
-        - Here are displayed all the pictures of the selected category.\n\n
-        - A representative picture is a picture that is used to represent the category in the dataset.
-        Such a picture is not part of the pictures of the category.\n\n
-        Possible actions :\n\n
-        - To delete pictures : click longer on the picture you want to delete, select all the pictures to delete and press the trashcan icon.\n\n
-        - To set an picture as representative of the category : click longer on the picture and select the star icon.</string>
```
bbjubjub2494, line 56: where did those go?

Jaakik: I just removed these when resolving the conflicts, you can see those same resources refactored just above

bbjubjub2494: Ok found it. It's different than the original tho, why is that? And where is the information on deleting a category and renaming a dataset?

bbjubjub2494: Looking at the view in context, these instructions seem redundant. Is that the idea?

Jaakik: Yes that was the idea initially but I think it's Ok to leave them for first time users so I ll put them back -- and readd the info that I mistakenly removed

bbjubjub2494 (approved):

